Issue #45, opened 2022-07-01
Operator Compare crash when Mode too large


hollyhook:
Connect the Compare operator input Mode with an int > 4 (?) causes a crash

```
System.ArgumentOutOfRangeException
  HResult=0x80131502
  Message=Specified argument was out of the range of valid values.
  Source=Operators
  StackTrace:
   at T3.Operators.Types.Id_026869ee_b62f_481e_aadf_f8a1db77fe65.Compare.Update(EvaluationContext context) in C:\dev\repos\t3\Operators\Types\Compare.cs:line 37
   at T3.Core.Operator.Slots.Slot' 1.Update(EvaluationContext context) in C:\dev\repos\t3\Core\Operator\Slots\Slot.cs:line 73
   at T3.Gui.OutputUi.OutputUi' 1.Recompute(ISlot slot, EvaluationContext context) in C:\dev\repos\t3\t3\Gui\OutputUi\OutputUi.cs:line 45
   at T3.Gui.OutputUi.OutputUi' 1.DrawValue(ISlot slot, EvaluationContext context, Boolean recompute) in C:\dev\repos\t3\t3\Gui\OutputUi\OutputUi.cs:line 31
   at T3.Gui.Windows.Output.OutputWindow.DrawOutput(Instance instanceForOutput, Instance instanceForEvaluation) in C:\dev\repos\t3\t3\Gui\Windows\Output\OutputWindow.cs:line 191
   at T3.Gui.Windows.Output.OutputWindow.DrawContent() in C:\dev\repos\t3\t3\Gui\Windows\Output\OutputWindow.cs:line 73
   at T3.Gui.Windows.Window.DrawOneInstance() in C:\dev\repos\t3\t3\Gui\Windows\Window.cs:line 93
   at T3.Gui.Windows.Output.OutputWindow.DrawAllInstances() in C:\dev\repos\t3\t3\Gui\Windows\Output\OutputWindow.cs:line 40
   at T3.Gui.Windows.Window.Draw() in C:\dev\repos\t3\t3\Gui\Windows\Window.cs:line 49
   at T3.Gui.Windows.WindowManager.Draw() in C:\dev\repos\t3\t3\Gui\Windows\WindowManager.cs:line 94
   at T3.Gui.T3Ui.Draw() in C:\dev\repos\t3\t3\Gui\T3UI.cs:line 92
[...]

PR #51, opened 2022-07-02
clamp input enums, to prevent IllegalArgumentException. Fix for #45

hollyhook:


pixtur:
Thanks! Added to next release!

PR #62, opened 2022-07-10
Reflection for types

mrvux:
Type registration improvements (uses reflection so we can dynamically register types/uitypes)
Symbol load by extension


mrvux:
Reopened, was set in the wrong branch.

pixtur:
Thank you!

PR #92, opened 2022-09-04
[Fix] Register port group properly in the dictionary when created.

### LINKED ISSUE #87: OSCInput - Multi address support
OSCInput only supports one address on the same port.
t3_eM1QtUqkEW
The signal from additional OSCInput nodes isn't recognized.

mrvux:
Was preventing to create multiple OSC Input instances
fixes #87


pixtur:
Thanks!

PR #101, opened 2022-09-13
Osc, small optimization

mrvux:
Directly cast the packet into a message or a bundle instead of using tostring and reparsing (also saves quite some gc)


pixtur:
I'm not sure how to merge this pull request without fixing the crash on change OSC port... :(

PR #104, opened 2022-09-16
Nested time clip zooming without local time handling

HolgerFoerterer:
Removed nested time handling completely - now, global scaling and scrolling get adjusted as you jump in or out of a timeline.
Also renamed InverseTransformPosition to InverseTransformPositionFloat for clarity.
Changed some scaling to float (TransformPositionFloat) instead of integer (TransformPosition).
Was it intended to have so many transformations running with integer coordinates? Should we cross-check the use?
You should now only need to override TransformPositionFloat and InverseTransformPositionFloat to get a functioning subclass.
Snapping works better now but snapping while pressing alt needs to be looked at.
Known bugs: The UI does not render correctly when rendering the timeline backwards (negative scale).

In addition: hovering over the output camera allows you to zoom in/out by using the center mouse wheel. No need to activate the window anymore.

### DIFF
--- a/T3/Gui/Graph/AnnotationElement.cs
+++ b/T3/Gui/Graph/AnnotationElement.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using ImGuiNET;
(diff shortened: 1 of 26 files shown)


pixtur:
Thanks!
> Also renamed InverseTransformPosition to InverseTransformPositionFloat for clarity.
To be honest, I'm not really sure about this change. I can see how the override can be misleading. But type function type overrides are a core features of c# and once with start doing this, we also might want to start to rename ` Remap()` because they all contain type overrides.

> Was it intended to have so many transformations running with integer coordinates? Should we cross-check the use?
Yes. Especially in the timeline it's important the keyframes sit precisely on the CurrentTimeMarker. 

> You should now only need to override TransformPositionFloat and InverseTransformPositionFloat to get a functioning subclass.
Okay. I will try to look into your implementation.

> In addition: hovering over the output camera allows you to zoom in/out by using the center mouse wheel. No need to activate the window anymore.
Awesome. But ideally this would have been two pull requests, I guess.

PR #110, opened 2022-09-17
Save files with new folder structure of operators in zip backup

hollyhook:
Saves files incl. folder names in backup zip for all below Operator\Types. Fixes issue reported here https://discord.com/channels/823853172619083816/986939763992981524/1020641989282250782

### DIFF
--- a/T3/Gui/AutoBackup/AutoBackup.cs
+++ b/T3/Gui/AutoBackup/AutoBackup.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
 using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


pixtur:
Hi @hollyhook!

Thank for your fix. Although I'm certain that would work, I believe it could implemented without recursion. (Which I try to avoid for sanity reasons).

Did you know the  `SearchOption.AllDirectories` options of `GetDirectories()` :

https://learn.microsoft.com/en-us/dotnet/api/system.io.searchoption?view=net-6.0

It returns a flattened recursive list and I'm using it extensively for the namespace directories implementation.

Issue #109, opened 2022-09-17
Crash in Gradient.cs


hollyhook:
One-time occurance, Win 11, crash happened while rendering w/o interaction

```
System.InvalidOperationException
  HResult=0x80131509
  Message=Collection was modified; enumeration operation may not execute.
  Source=System.Private.CoreLib
  StackTrace:
   at System.ThrowHelper.ThrowInvalidOperationException_InvalidOperation_EnumFailedVersion()
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at T3.Core.DataTypes.Gradient.Write(JsonTextWriter writer) in C:\dev\repos\t3\Core\DataTypes\Gradient.cs:line 23
   at T3.Core.Model.<>c.<.ctor>b__3_34(JsonTextWriter writer, Object obj) in C:\dev\repos\t3\Core\Resource\Model.cs:line 255
   at T3.Core.Operator.InputValue`1.ToJson(JsonTextWriter writer) in C:\dev\repos\t3\Core\Operator\Slots\InputValue.cs:line 76
   at T3.Core.SymbolJson.WriteSymbolChildren(List`1 children) in C:\dev\repos\t3\Core\Resource\SymbolJson.cs:line 120
   at T3.Core.SymbolJson.WriteSymbol(Symbol symbol) in C:\dev\repos\t3\Core\Resource\SymbolJson.cs:line 49
   at T3.Core.Model.SaveSymbolDefinitionAndSourceFiles(IEnumerable`1 valueCollection) in C:\dev\repos\t3\Core\Resource\Model.cs:line 541
   at T3.Gui.UiModel.SaveModifiedSymbols() in C:\dev\repos\t3\t3\Gui\UiModel.cs:line 311
   at T3.Gui.T3Ui.SaveModified() in C:\dev\repos\t3\t3\Gui\T3UI.cs:line 271
   at T3.Gui.AutoBackup.AutoBackup.CreateBackupCallback() in C:\dev\repos\t3\t3\Gui\AutoBackup\AutoBackup.cs:line 49
   at System.Threading.Tasks.Task.InnerInvoke()
   at System.Threading.Tasks.Task.<>c.<.cctor>b__277_0(Object obj)
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)

```

hollyhook:
Crash just occured again.
[...]

PR #119, opened 2022-09-20
Fix for issue #109 

domportera:
Basically just made a data copy of the Steps list before trying to write with the data

also this is my first-ever FOSS pull request so that means you have to accept it

### DIFF
--- a/Core/DataTypes/Gradient.cs
+++ b/Core/DataTypes/Gradient.cs
@@ -20,7 +20,14 @@ namespace T3.Core.DataTypes
             writer.WritePropertyName("Steps");
             writer.WriteStartArray();
 
-            foreach (var step in Steps)
+            StepStruct[] stepsToWrite;
+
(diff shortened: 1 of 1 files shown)


pixtur:
Hi @domportera!

thanks for your pull-request. Although I believe that this would work, replicating the `Step` class as a struct seems like a dirty work around that will hunt us as soon as somebody tries to extend the original class.

I think the idea of cloning the list before serializing is great, but it would be better to extend the original class with `Clone` method to do so.

domportera:
Fair, I considered that myself but thought it was a mild compromise. The thought was that using structs during this serialization step, there's no chance that anyone could modify the original object while it's being serialized. That being said I'll modify it to duplicate the objects into instances of their own class and submit a new pull request when I get the chance :)

pixtur:
Awesome. Thank you!

PR #124, opened 2022-09-23
Fix for issue #109 take II: the sequel

domportera:
sorry this PR accidentally includes something that's not in the dev branch, if you reject we can make it a trilogy and i wont mind but i figure it's just a template


pixtur:
Excellent! Looks much better.

Issue #98, opened 2022-09-10
[Bug] Multi output types not fully disabled with Shift+D


VeissieV:
When disabling types with multiple outputs with Shift+D the additional outputs are not disabled. Tested with Commands and Textures.
![unknown](https://user-images.githubusercontent.com/26267774/189461494-89aefb66-31ab-43c4-a807-2d05aac170ca.png)

pixtur:
Thanks for reporting. This should be fixed.

domportera:
claiming this one

pixtur:
Need for verify if this is still valid.

PR #135, opened 2022-09-25
Multi-output nodes properly disabled

domportera:
I may have fixed issue #98.

Not gonna lie though, I don't know much about how these systems work so please let me know if my comments are representative of the intended functionality or if this will produce unwanted functionality. I should really read the wiki and learn to properly use the program

Also what's your feeling about logging warnings/errors? I threw those in because I know I would want them on my own project. There were null checks there already and I imagine if something there is null then something has gone wrong and we would want to know about it.


pixtur:
Strange. For some reason the default file encoding changed. Please check your editor:
image

T3/Gui/SymbolChildUi.cs
```diff
@@ -1,14 +1,18 @@
...
+using Microsoft.VisualBasic.Logging;
 using T3.Core.Operator;
 using T3.Gui.ChildUi;
 using T3.Gui.Graph;
 using T3.Gui.Graph.Interaction;
 using T3.Gui.Selection;
 using UiHelpers;
+using T3.Core.Logging;
+using Log = T3.Core.Logging.Log;
+using T3.Core.Operator.Slots;
```
pixtur, line 15:
```suggestion
using System;
using System.Collections.Generic;
using System.Linq;
using System.Numerics;
using ImGuiNET;
using T3.Core.Logging;
using T3.Core.Operator;
using T3.Gui.ChildUi;
using T3.Gui.Graph.Interaction;
using T3.Gui.Selection;
using UiHelpers;
using T3.Core.Operator.Slots;
```

T3/Gui/SymbolChildUi.cs
```diff
@@ -46,19 +50,49 @@ public enum ConnectionStyles
...
+                    {
+                        Log.Warning($"{typeof(SymbolChild)} {SymbolChild.ReadableName} does not have the following child output as defined: " +
+                            $"{childOutput.OutputDefinition.Name}({nameof(Guid)}{childOutput.OutputDefinition.Id})");
+                        continue;
+                    }
+
                     childOutput.IsDisabled = value;
                 }
 
+                //Set disabled status on outputs of each instanced copy of this child within all parents that contain it
```
pixtur, line 77:
```suggestion
                // Set disabled status on outputs of each instanced copy of this child within all parents that contain it
```

pixtur, line 77:
```suggestion
                // Set disabled status on outputs of each instanced copy of this child within all parents that contain it
```

T3/Gui/SymbolChildUi.cs
```diff
@@ -46,19 +50,49 @@ public enum ConnectionStyles
             get => SymbolChild.Outputs.FirstOrDefault().Value?.IsDisabled ?? false;
             set 
             {
-                var firstOutputDef = SymbolChild.Symbol.OutputDefinitions.FirstOrDefault();
-                if (firstOutputDef != null)
+                List<Symbol.OutputDefinition> outputDefinitions = SymbolChild.Symbol.OutputDefinitions;
+
+                //Set disabled status on this child's outputs
```
pixtur, line 55:
```suggestion
                // Set disabled status on this child's outputs
```

pixtur (changes requested):
The code looks fine, but I commented some c# style / formatting changes

domportera:
@pixtur I'll look into that text encoding thing - what text encoding are you using?

and your formatting preferences are noted! will do my best to abide by them next time and be more vigilant about checking the namespaces that VS is automatically adding

defaulting to `foreach` works for me and makes my life easier and the code more readable. im not a big fan of the `var` keyword in many circumstances but I can definitely adopt your style here.

FWIW (not much even to me), I often opt for `for` for the sake of the occasional (however negligible) performance boost depending on the collection size and type.

pixtur:
> @pixtur I'll look into that text encoding thing - what text encoding are you using?
I still don't know what's causing this. My Rider IDE tells me this: 
image

> FWIW (not much even to me), I often opt for for for the sake of the occasional (however negligible) performance boost depending on the collection size and type.

Are you sure about this? I would even argue the opposite: With for each you can frequently use IEnumerable instead of allocating lists. (Linq is another topic altogether. I'm trying to fix the heavy GC hitters one at a time.) If in doubt I'm with [Jon Skeet](https://stackoverflow.com/a/472258) on that one.

pixtur:
Merged to dev

domportera:
yes it seems like he agrees with my premise that for Lists a for loop is considerably faster (though this advantage is disappearing with .NET 7!). with Arrays there is no difference performance wise.

I don't think any heap memory is allocated in the process, just with foreach Lists implement IEnumerable enumeration, while in a for loop you can force the use of the faster array indexer. no additional array or list is created in the process, it's just using the underlying array structure already there

I could be wrong but I nerded out about this recently and feel that's accurate

pixtur:
I don't consider myself to be good programmer. But I love to discuss stuff like this. Sometimes it makes a difference so it's always a good idea to have a rough understand how your code might be translated by the compiler. For List vs. Array I'm actually surprised that the c# compiler isn't able to optimize that.

I personally always have to smile when seeing code like `for(int i=0; i<100; ++i) {...} ` 
I know why some people think it makes a difference (or used to make). But frequently these people are also occupied with micro performance optimization instead of shipping a great product. 

Like with every interesting question: It depends. And the truth lies often somewhere in middle.

domportera:
I consider you to be a great programmer :) but yeah this stuff is fascinating despite its relative lack of utility. I was surprised too - i was generally a foreach-til-death guy until I learned that recently. but with .NET 7 coming around fortunately it's not a habit i'll really have to kick and it's convenient that you're not concerned

I feel that. I have a friend who insisted on not creating stack variables (for readability) for "optimization" purposes. times have changed thankfully 😂 (unless you're working with shaders... but we dont talk about that)

Issue #121, opened 2022-09-21
[DampVec2] and [DampVec3]


pixtur:
It would be great if [Damp] would be available for vector types.

VeissieV:
How do you mean? I made a type that applies the same dampening to XYZ and another that allows for independent damping for each.

<img width="1263" alt="t3_PSkLaawyqa" src="https://user-images.githubusercontent.com/26267774/191630326-f7f48a0f-7a05-4ebe-b115-a7111f37d20a.png">

pixtur:
Yes. Precisely that.  But I would do this in c#. Duplicating the [Damp] operator and converting the float parameter types to Vector2  and 3 should be straight forward. Building this from Operator components has an unnecessary performance overhead.

pixtur:
Fixed in next release

PR #137, opened 2022-09-25
some refactoring and utility functions to make extending Damp to other types simpler

domportera:
for #121

### DIFF
--- a/Core/Resource/MathUtils.cs
+++ b/Core/Resource/MathUtils.cs
@@ -285,66 +285,25 @@ namespace T3.Core
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


pixtur:
This looks quite good! Then only hesitation is the ` CheckNaN()`  and `CheckInfinity()` methods.
I'm not sure if the the purpose justifies a helper method -- but I can totally see how it can clarify the code.

How about combining both into something like `ApplyDefaultIfInvalid()` ?

domportera:
sounds good to me! I'll get on that as soon as I get the chance

domportera:
done!

PR #141, opened 2022-09-25
Log mesh u vs

domportera:
for #4 
I don't think this covers every edge case regarding multiple UV sets and I'm making some assumptions on how mesh UVs are handled here, but I figured I'd clean up some low-hanging fruit if I could and put it somewhere it could be easily maintainable.

### DIFF
--- a/Operators/Types/lib/3d/mesh/LoadObj.cs
+++ b/Operators/Types/lib/3d/mesh/LoadObj.cs
@@ -42,7 +42,7 @@ namespace T3.Operators.Types.Id_be52b670_9749_4c0d_89f0_d8b101395227
             if (path != _lastFilePath || SortVertices.DirtyFlag.IsDirty || Math.Abs(scaleFactor - _scaleFactor) > 0.001f)
             {
                 _scaleFactor = scaleFactor;
-                
+
                 _description = System.IO.Path.GetFileName(path);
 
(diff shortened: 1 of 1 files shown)


pixtur:
Hi @domportera ,

this was bad timing indeed! You code looks fine and is valid, but I was simultaneously working on [LoadObj] and `ObjMesh` to overcome this issue: Now when loading Obj-files without UV coordinates we generate a fallback UV-setup via box projection:

https://github.com/still-scene/t3/commit/446e8ab41b930c7b6af73f3c83666b168927a10e
https://github.com/still-scene/t3/commit/cb63b4ad5dace9ce34114d89f35209ab81fc1fe1

So I believe these warnings are no longer required. 

Maybe we should align to avoid such conflicts. Depending on what you want to work on, it might be good to go through the [RoadMap](https://github.com/orgs/still-scene/projects/1/views/2) 

I tagged some tasks [here](https://github.com/orgs/still-scene/projects/1/views/1?filterQuery=label%3A%22help+wanted%22&visibleFields=%5B%22Title%22%2C%22Assignees%22%2C%22Status%22%2C2957806%2C%22Labels%22%5D)

domportera:
no worries! that sounds like a *much* better solution. fortunately this was a super quick thing that just helped me learn the codebase a little more. I'll take a look!

PR #157, opened 2022-09-28
Video output with LoopRange, Bars and Frames

HolgerFoerterer:
I cleaned up the render sequence and render video output:

* You can now choose to use the LoopRange as render range
* You can select between Bars, Seconds and Frames as time reference
* I cleaned up the buffering to get frames at precise times (incoming textures are normally *two* frames behind)
* The new class RenderHelperWindow is used to group GUI and values that are used both for sequence and video output
* Changing settings for the sequence output will change settings for the video output and vice versa
* Some more bug fixes

I tested frame skipping/shifting with an NVIDIA card. It would be interesting to know if it works on AMD graphics cards in exactly the same way.

* *Known bug:* Using two enum dropdown boxes proved to be impossible. They would show wrong or no entries. I now used a temporary fix in T3/Gui/Styling/CustomComponents.cs to be able to present the PNG/JPG selection option in T3/Gui/Windows/RenderSequenceWindow.cs, but it's a hack and should be replaced by a proper solution, for example with instance numbering.
* *Known Bug:* Disabling double-buffering in the driver may cause the renderings to be off by one or two frames again
* *Known Bug:* MP4 video output still does not allow non-integer fps

### DIFF
--- a/Core/Animation/Playback.cs
+++ b/Core/Animation/Playback.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Diagnostics;
 
 namespace T3.Core.Animation
@@ -93,12 +93,12 @@ namespace T3.Core.Animation
 
         public double BarsFromSeconds(double secs)
         {
-            return secs * Bpm / 240f;
+            return secs * Bpm / 240.0;
         }
         
         public double SecondsFromBars(double bars)
         {
-            return bars * 240 / Bpm;
+            return bars * 240.0 / Bpm;
         }
         
         private static double _lastFrameStart;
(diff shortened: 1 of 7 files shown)


Core/Animation/Playback.cs
```diff
@@ -93,12 +93,12 @@ public virtual void Update(bool idleMotionEnabled = false)
...
         public double BarsFromSeconds(double secs)
         {
-            return secs * Bpm / 240f;
+            return secs * Bpm / 240.0;
         }
         
         public double SecondsFromBars(double bars)
         {
-            return bars * 240 / Bpm;
+            return bars * 240.0 / Bpm;
```
pixtur, line 101: Good catch! Although Bpm should be in double, this makes the code much more readable.

pixtur (approved):
Stellar contribution! Thanks a lot.

Issue #133, opened 2022-09-23
Some Operators are in incorrect location or namespace


hollyhook:
We made a list of operators which are potentially in the wrong folder or namespace, check out
https://github.com/still-scene/t3/wiki/Operator-in-incorrect-location-or-namespace

@pixtur please correct and/or complete where the right location would be

I'm happy to do the cleanup and create a PR

pixtur:
@hollyhook I agree with all of you suggestions. Can you move them (either with the edit field in the parameter window or by dragging them with the symbol browser) to the correct locations and submit a pull request?

hollyhook:
In some cases I don't know where to put the Operator.
@pixtur can you please look into the list and clarify the things marked with "?"

pixtur:
Ups. Just did so!

hollyhook:
done

PR #160, opened 2022-09-28
Some operators moved to a better matching folder

hollyhook:
As discussed here: https://github.com/still-scene/t3/wiki/Operator-in-incorrect-location-or-namespace.

Fix for #133


pixtur:
Thanks!

Issue #67, opened 2022-07-17
Outputs can't be removed with Delete key


hollyhook:
- Go into an operator
- Select any output by clicking with the mouse on it, so that it's highlighted
- try to remove it with Delete key 
 -> Selection vanishes but the output is not removed

Removing via context menu (Right click -> Delete) works as expected

pixtur:
I can reproduce this issue. Thanks for reporting.

domportera:
this was resolved in #171, can we close this?

pixtur:
I checked and found some more issues related to this topic:
- Undo doesn't work
- Deleting output from code-ops caused exceptions -> Fixed
- Deleting doesn't work if output has code overrides like Transform-Gizmos

PR #171, opened 2022-10-01
#67 delete key and delete menu functionality consistency

domportera:
#67


domportera:
love it, forgot about that one! i can do a little PR for this later if you haven't done it already

Issue #123, opened 2022-09-21
Invalid number input should not result in NaN


pixtur:
**To Reproduce**
1. edit parameter
2. type 1/0 
-> value is now NaN

**Expected behavior**
Parameter should revert to default value or 0

pixtur:
Fixed in next release

PR #172, opened 2022-10-01
Nan property handling

domportera:
this took way longer to find than i'm proud to admit

for #123 

and apparently #177

### DIFF
--- a/T3/Gui/Interaction/SingleValueEdit.cs
+++ b/T3/Gui/Interaction/SingleValueEdit.cs
@@ -168,7 +168,15 @@ namespace T3.Gui.Interaction
                                 _editValue = _startValue;
                         }
 
-                        _editValue = Evaluate(_jogDialText);
+                        bool invalid;
+                        _editValue = Evaluate(_jogDialText, out invalid);
+
+                        //if the value is invalid, just revert it to what it was previously
+                        if (invalid)
+                        {
+                            _editValue = _startValue;
+                            _jogDialText = value.ToString();
+                        }
                         break;
                 }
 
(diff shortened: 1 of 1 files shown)


pixtur:
Thanks, this is already a great fix! 

One thing we could improve in the long term is to actually use the defined default value of the edited property.

This might be possible by returning 

```
    public enum InputEditStateFlags
    {
        Nothing = 0x0,
        Started = 0x1,
        Modified = 0x2,
        Finished = 0x4,
        ShowOptions = 0x8,
        ResetToDefault = 0x10,
        ModifiedAndFinished = Modified | Finished
    }
```
`InputEditStateFlags.ResetToDefault`  

I will quickly look into this.

domportera:
Thanks! It took a few iterations to find the real *root* of the problem but I'm glad I found it.

I think I tried ResetToDefault and didn't have luck, but im sure I was doing something wrong. I'm sure you'll have better luck than I did :)

T3/Gui/Interaction/SingleValueEdit.cs
```diff
@@ -390,6 +401,7 @@ private enum InputStates
 
         private static string _numberFormat = "{0:0.000}";
         private static double _timeOpened;
+        private const double DefaultValue = 0.0;
```
pixtur, line 404: should be `_defaultValue` because of private

pixtur:
You code looks really nice and works. I'm really impressed how quickly you were able to dive into that.

T3/Gui/Interaction/SingleValueEdit.cs
domportera: noted!

domportera:
thanks!! i do my best :) just took a lot of hunting and a strong magnifying glass 🕵️‍♂️

PR #173, opened 2022-10-01
Feat/borderless mouse

DartFNM:
A small but useful feature. 
Allows the mouse cursor to keep moving when it reaches the edge of the screen and to turn the window around without interrupting parameter changes.

borderless


Install/default-layouts/layout0.json
```diff
@@ -4,35 +4,35 @@
...
     "Size": {
       "X": 0.553125,
-      "Y": 0.98447424
+      "Y": 0.9840278
     }
   },
   {
     "Title": "Settings",
     "Visible": false,
     "Position": {
```
pixtur, line 17: Please avoid committing files that are not relevant for actual change.
This will make reviewing easier and avoid side effects and regressions.

pixtur: Ah. i think these changes were caused by git branching interference. I can clean this up locally.

T3/App/ImGuiDx11RenderForm.cs
```diff
@@ -17,7 +17,7 @@ public class ImGuiDx11RenderForm : RenderForm
         public ImGuiDx11RenderForm(string title)
             : base(title)
         {
-            MouseMove += (o, e) => ImGui.GetIO().MousePos = new System.Numerics.Vector2(e.X, e.Y);
+            //MouseMove += (o, e) => ImGui.GetIO().MousePos = new System.Numerics.Vector2(e.X, e.Y);
```
pixtur, line 20: Please avoid commented code without explaining comment while it's still relevant.
In this case removing it would have been just fine.

T3/App/ImGuiDx11RenderForm.cs
```diff
@@ -76,10 +77,15 @@ protected override void WndProc(ref System.Windows.Forms.Message m)
                     if (m.Msg == WM_LBUTTONDOWN || m.Msg == WM_LBUTTONDBLCLK) button = 0;
                     if (m.Msg == WM_RBUTTONDOWN || m.Msg == WM_RBUTTONDBLCLK) button = 1;
                     if (m.Msg == WM_MBUTTONDOWN || m.Msg == WM_MBUTTONDBLCLK) button = 2;
-                    // TODO
+                    io.MouseDown[button] = true;
+
                     //if (!ImGui.IsAnyMouseDown() && ::GetCapture() == NULL)
```
pixtur, line 82: Speaking or commented out code... 🙈

T3/App/ImGuiDx11RenderForm.cs
```diff
@@ -91,11 +97,48 @@ protected override void WndProc(ref System.Windows.Forms.Message m)
...
+                            int rightEdge = ClientRectangle.Right - 12;
+                            int leftEdge = ClientRectangle.Left + 12;
+
+                            bool wrapped = false;
+                            if (xPos > rightEdge)
+                            {
+                                xPos = leftEdge;
+                                wrapped = true;
+                            }
+                            if (xPos < leftEdge)
```
pixtur, line 129:
```suggestion
                            else if (xPos < leftEdge)

```

T3/App/ImGuiDx11RenderForm.cs
```diff
@@ -91,11 +97,48 @@ protected override void WndProc(ref System.Windows.Forms.Message m)
...
+                            {
+                                xPos = rightEdge;
+                                wrapped = true;
+                            }
```
pixtur, line 133: I would leave an empty line here.

T3/Program.cs
```diff
@@ -37,7 +37,7 @@ public static class Program
         public static SpaceMouse SpaceMouse { get; private set; }
 
         public static bool IsStandAlone = File.Exists("StartT3.exe");
-        public const string Version = "v3.3.0";
+        public const string Version = "v3.4.0";
```
pixtur, line 40: Code catch. But this would have justified it's own PR :D

T3/Gui/Interaction/SliderLadder.cs
```diff
@@ -56,10 +56,14 @@ public static void Draw(ref double editValue, ImGuiIOPtr io, double min, double
...
-            var pNow = io.MousePos - center;
+            Vector2 pLast = io.MousePos - io.MouseDelta - center;
+            Vector2 pNow = io.MousePos - center;
+            if (!ImGui.IsMousePosValid())
+            {
+                _lastStepPosX = pNow.X = float.NaN;
+            }
 
-            if (timeSinceVisible < 0.032f)
+            if (timeSinceVisible < 0.032f || float.IsNaN(_lastStepPosX) )
```
pixtur, line 66: This looks kind of dirty. 
Why not just use a boolean "isDirty" instead of setting `_lastStepPos`?

T3/Gui/Interaction/SliderLadder.cs
```diff
@@ -109,7 +113,12 @@ public static void Draw(ref double editValue, ImGuiIOPtr io, double min, double
                                    range.Label);
             }
 
-            var deltaSinceLastStep = pLast.X - _lastStepPosX;
+            float deltaSinceLastStep = 0;
+            if (!float.IsNaN(_lastStepPosX))
+            {
+                deltaSinceLastStep = pLast.X - _lastStepPosX;
+            }
+
```
pixtur, line 121:
```suggestion
            float deltaSinceLastStep =  float.IsNaN(_lastStepPosX)
                                                 ? 0 
                                                 : pLast.X - _lastStepPosX;
```

pixtur (changes requested):
I think this is a great contribution.
However it will definitely interfere with other drag manipulations and interactions.

After testing this I believe the feature needs to be more polished to become a default.

Some things that come to mind: 
- some components (like the value ladder or dragging the canvas background) could register a WrappedMouseCapture state (e.g. by setting a global static variable in T3UI) so we can limit the side effects on other components.
- We could also check if the current input method is a mouse or stylus and disable wrapping for that.
- It might also be interesting to check if we could catch the mouse point in place after reaching the lock threshold of the value ladder. This would allow users to still switch between scale regions but would avoid the irritation from the wrapping. (Even after a couple of years of blender, this still irritates me once in a while).

What do you think?

Here is a quick test capture:

https://user-images.githubusercontent.com/1732545/193634882-047141fd-0f3f-4bc1-92cb-87029feb2951.mp4

pixtur:
Closing this for now. We can reopen it, as soon as we start working on this again.

PR #175, opened 2022-10-01
Feat/autorepeat rewind timeline

DartFNM:
Auto-repeat rewind timeline action from:
`KeyboardBinding.Triggered(UserActions.PlaybackNextFrame)`

Now the timeline moves automatically when Shift+Left\RightArrowKey is pressed

autorepeat


T3/Gui/Interaction/UserAction.cs
```diff
@@ -148,19 +151,55 @@ public static bool Triggered(UserActions action)
...
+                        && ((!c.Shift && !io.KeyShift) || (c.Shift && io.KeyShift))
+                        )
+                    {
+                        if (binding.PressTime == 0) // first time key down
+                        {
+                            binding.PressTime = curTime;
+                            return true;
+                        }
 
+                        float deltaTime = curTime - binding.PressTime;
```
pixtur, line 182: I would have called this `timeSinceKeyDown` or something like this.

T3/Gui/Interaction/UserAction.cs
```diff
@@ -126,6 +126,9 @@ public class KeyboardBinding
         public bool NeedsWindowHover;
         public bool KeyPressOnly;
         public readonly KeyCombination Combination;
+        public float AutoRepeatTime = -1f;
+        private float PressTime = 0;
```
pixtur, line 130: I believe, this is not optimal: The KeyboardBinding class should be considered "immutable" without any state. It's basically a hotkey definition. As soon as you add state that depends on user interaction, things start to get messy.

I can see however, how this is a tricky case because the "state" is being held by ImGui.

Are you sure, you can't use any of the ImGui features likes...

```
    // uses provided repeat rate/delay. return a count, most often 0 or 1 but might be >1 if RepeatRate is small enough that DeltaTime > RepeatRate
    ImGui.GetKeyPressedAmount(ImGuiKey key, float repeat_delay, float rate);  
```

pixtur (changes requested):
I would love to have this PR. Constantly pressing Shift+Left/Right drives me nuts.
But I also believe the code is not there yet.

I believe we should find another option to get the press duration of a key. I'm convinced that ImGui already provides that feature somehow.

If not, we can find another solution that changes this "hack" into a feature that can used by other components.

pixtur:
Closing this for now.

Issue #156, opened 2022-09-27
Layout of settings window should be improved


pixtur:
The current layout is confusing and hard to read:

We should...

- remove obsolete options and debug output
- switch to left sided labels 

![image](https://user-images.githubusercontent.com/1732545/192649562-43b7ac86-23f4-4990-9ebb-0eaf018ff757.png)

domportera:
Does this look like I'm on the right track?

![image](https://user-images.githubusercontent.com/6530600/193446329-9989d242-3124-4b81-b081-83f2c026cba0.png)

pixtur:
Nice!

Some comments:
- Are you using the `CustomComponents.FloatValueEdit()` ? You can check how this is used in `RenderVideoWindow.cs`.
- The Checkbox should stay on the left side. (I have a kind of trauma on that topic :) )

PR #178, opened 2022-10-02
Improved settings UI

domportera:
for #156

### DIFF
--- a/T3/Gui/SettingsUi.cs
+++ b/T3/Gui/SettingsUi.cs
@@ -1,0 +1,64 @@
+using ImGuiNET;
+using System;
+using System.Collections.Generic;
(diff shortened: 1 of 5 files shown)


pixtur (changes requested):
Very interesting, that's a very interesting solution.
I can see how a more thorough definition of a user setting (e.g. cobined with label, default value and tooltip, etc.) could be valuable in the long term.

For now however, I would stick with the ImGui way of doing things: I was thinking about just switching from `ImGui.DrawFloat()` to `CustomControls.DragFloat()`. This would basically do the same thing with 200 lines less code.

domportera:
closing this to open a new pull request with a new branch bc i kinda suck at rebasing

Issue #162, opened 2022-09-28
Entering a selected operator and dragging child elements will drag parent operator


pixtur:
**Please complete the following information):**
 - OS: win 10
 - Version: v3.4

**To Reproduce**
1. Select operator
2. Double click to enter
3. Drag a child operator
4. Double click on background to exit to parent

-> Parent operator has been moved

**Expected behavior**
Parent operator should not be moved.

**Additional context**
I the drag interaction probably isn't cleared when entering the sub patch.

This bug is very destructive because move operators are not immediately noticeable.

PR #179, opened 2022-10-03
Dragging a child node no longer affects parent

domportera:
fix for #162 plus some housekeeping

[a little housekeeping](https://github.com/still-scene/t3/commit/0b3a396ea0542b96ea87e6a31968ff920815528a) turned a // TODO into a // TODID. I probs shouldve branched separately but i guess i felt lazy

### DIFF
--- a/T3/Gui/Graph/GraphNode.cs
+++ b/T3/Gui/Graph/GraphNode.cs
@@ -35,6 +35,12 @@ namespace T3.Gui.Graph
     {
         public static void Draw(SymbolChildUi childUi, Instance instance)
         {
+            if (_instanceToSwitchGraphWindowTo != null)
+            {
+                GraphCanvas.Current.SetCompositionToChildInstance(_instanceToSwitchGraphWindowTo);
+                _instanceToSwitchGraphWindowTo = null;
+            }
+
             if (instance == null)
                 return;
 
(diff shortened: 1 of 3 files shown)


pixtur (commented):
Great work! I'm convinced this would fix the issue.

However, I'm a slightly reluctant to implement the fix on such a low level:
`SelectableNodeMovement.Handle` is called for every op on every frame. For this issue it might have been sufficient to reset SelectableNovement from `GraphCanvas.SetComposition() {...}`.


I looked into this and found that this solutions seems to work:

In `SelectableNodeMovement.cs` ...
```
        /// <summary>
        /// Reset to avoid accidental dragging of previous elements 
        /// </summary>
        public static void Reset()
        {
            _moveCommand = null;
            _draggedNodes.Clear();
        }
```        

In `GraphCanvas.cs` ...


```
        public void SetComposition(List<Guid> childIdPath, ICanvas.Transition transition)
        {
            SelectableNodeMovement.Reset();
            
```

I guess this is one of these cases where it just takes very long to understand a complex problem so I can be fixed by something really simple.

domportera:
>is called for every op on every frame
OH

yes your solution is *much* better. I was trying to keep `GraphCanvas`'s grubby little hands out of the selection handling where possible, but knowing that changes things

PR #180, opened 2022-10-03
New op PickColorFromImage

hollyhook:
I submit this PR despite the code having at least one major issue, hoping to get feedback how to fix it.
Issue is, I don't know how to find out how many bits a pixel has., even after extensive stackoverflow search. Right now, 32 bits are assumed, but that's wrong for many texture formats. Check out the code marked with FIXME.

An alternative fix would be to add [RenderTarget], and always enforce conversion to R8G8B8A8_UNORM format.

Also I really don't know what I'm doing, copy&pasted the code mostly from  https://github.com/still-scene/t3/blob/master/Operators/Types/user/cynic/research/Edtaa.cs

### DIFF
--- a/Operators/Types/user/meta/PickColorFromImage.cs
+++ b/Operators/Types/user/meta/PickColorFromImage.cs
@@ -1,0 +1,95 @@
+using System;
+using System.Numerics;
+using T3.Core;
+using T3.Core.Operator;
+using T3.Core.Operator.Attributes;
+using T3.Core.Operator.Slots;
+using SharpDX.Direct3D11;
+using SharpDX.DXGI;
+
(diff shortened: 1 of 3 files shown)


hollyhook:
Okay, i can then just support the formats 

- R16G16B16A16_Float
- R8G8B8A8_UNorm
- R16G16B16A16_UNorm

just like this https://github.com/still-scene/t3/blob/d5ff028fc7d71965abccb60b4760a1985d47bd03/T3/Gui/Windows/ScreenshotWriter.cs#L110?

pixtur:
I merged this to dev and added format conversions for 16bit.
I reused some code I found when I was working on the screenshot writer, so I can feel your pain. It took me several days to figure this out...

PR #192, opened 2022-10-08
Improved settings UI take two, SymbolBrowser improvements

domportera:
So I applied your feedback for the settings UI. ~~It's a little bit awkward imo (i still preferred the table ;))~~ I made it a table this time around too with that final commit. let me know if the way i did it though works a little better, i just found it unpleasant to work with without that

also made a little settings display class to make rendering lists of settings easier

Also made some significant changes to how the SymbolBrowser organizes itself - now it should never clip off of the canvas, and it should stay in view. You can now also be confident the description panel will stay up so you can scroll and pull out examples.

Also did some refactoring for maintainability and editability

sorry this pull request is gigantic lol

### DIFF
--- a/Core/Resource/MathUtils.cs
+++ b/Core/Resource/MathUtils.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Numerics;
+using System.Reflection;
 using T3.Core.Animation;
 using T3.Core.Operator;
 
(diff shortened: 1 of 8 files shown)


domportera:
closing for a squashed commit to make your life easier

PR #194, opened 2022-10-08
Improved Settings UI, consistent visibility and usability of SymbolBrowser, some refactoring

domportera:
So I applied your feedback for the settings UI. ~~It's a little bit awkward imo (i still preferred the table ;))~~ I made it a table this time around too with that final commit. let me know if the way i did it though works a little better, i just found it unpleasant to work with without that

also made a little settings display class to make rendering lists of settings easier

Also made some significant changes to how the SymbolBrowser organizes itself - now it should never clip off of the canvas, and it should stay in view. You can now also be confident the description panel will stay up so you can scroll and pull out examples.

Also did some refactoring for maintainability and editability

### DIFF
--- a/Core/Resource/MathUtils.cs
+++ b/Core/Resource/MathUtils.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Numerics;
+using System.Reflection;
 using T3.Core.Animation;
 using T3.Core.Operator;
 
@@ -144,6 +145,28 @@ namespace T3.Core
             }
 
             return v;
+        }
+
+        /// <summary>
+        /// A simple method to get a new index by "jumping" from a startIndex
+        /// When the jump would make the index exceed the valid range of the collection,
+        /// it will return an index that "wraps" around the other side
+        /// </summary>
+        /// <param name="startIndex">Index to jump from</param>
+        /// <param name="jumpAmount">How far to jump (usually 1 or -1 for forward and backward respectively)</param>
+        /// <param name="collection">The collection whose Count (Length for arrays) will be referenced</param>
+        /// <param name="realWrap">If false, this will default to making every wrap result in the first or last index of the collection.
+        /// If true, it will be a "real" wrap, where if you jump in excess of X, you will wrap to the index X distance from the other side. </param>
+        /// <returns></returns>
+        public static int WrapIndex(int startIndex, int jumpAmount, System.Collections.IList collection, bool realWrap = false)
+        {
+            int index = startIndex + jumpAmount;
+            index %= collection.Count;
+
+            bool newIndexIsNegative = index < 0;
+            int negativeWrapIndex = realWrap ? collection.Count + index : collection.Count - 1;
+
+            return newIndexIsNegative ? negativeWrapIndex : index;
         }
 
         public static Vector2 Min(Vector2 lhs, Vector2 rhs)
--- a/Core/Resource/MathUtils.cs
+++ b/Core/Resource/MathUtils.cs
@@ -1,8 +1,6 @@
 using System;
 using System.Numerics;
-using System.Reflection;
 using T3.Core.Animation;
-using T3.Core.Operator;
 
 namespace T3.Core
 {
@@ -145,28 +143,6 @@ namespace T3.Core
             }
 
             return v;
-        }
-
-        /// <summary>
-        /// A simple method to get a new index by "jumping" from a startIndex
-        /// When the jump would make the index exceed the valid range of the collection,
-        /// it will return an index that "wraps" around the other side
-        /// </summary>
-        /// <param name="startIndex">Index to jump from</param>
-        /// <param name="jumpAmount">How far to jump (usually 1 or -1 for forward and backward respectively)</param>
-        /// <param name="collection">The collection whose Count (Length for arrays) will be referenced</param>
-        /// <param name="realWrap">If false, this will default to making every wrap result in the first or last index of the collection.
-        /// If true, it will be a "real" wrap, where if you jump in excess of X, you will wrap to the index X distance from the other side. </param>
-        /// <returns></returns>
-        public static int WrapIndex(int startIndex, int jumpAmount, System.Collections.IList collection, bool realWrap = false)
-        {
-            int index = startIndex + jumpAmount;
-            index %= collection.Count;
-
-            bool newIndexIsNegative = index < 0;
-            int negativeWrapIndex = realWrap ? collection.Count + index : collection.Count - 1;
-
-            return newIndexIsNegative ? negativeWrapIndex : index;
         }
 
         public static Vector2 Min(Vector2 lhs, Vector2 rhs)
--- a/T3/Gui/Graph/Interaction/SymbolBrowser.cs
+++ b/T3/Gui/Graph/Interaction/SymbolBrowser.cs
@@ -343,17 +343,8 @@ namespace T3.Gui.Graph.Interaction
             var yOverflow = maxYPos > windowSize.Y;
             var yShrinkage = yOverflow ? windowSize.Y - maxYPos : 0;
 
-            position = new Vector2
-            {
-                X = position.X + xPositionOffset,
-                Y = position.Y
-            };
-
-            size = new Vector2
-            {
-                X = size.X,
-                Y = size.Y + yShrinkage
-            };
+            position.X += xPositionOffset;
+            size.Y += yShrinkage;
         }
 
         private void DrawDescriptionPanel(Vector2 position, Vector2 size, SymbolUi itemForHelp)
--- a/T3/Gui/Graph/Interaction/SymbolBrowser.cs
+++ b/T3/Gui/Graph/Interaction/SymbolBrowser.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Numerics;
+using System.Threading.Tasks;
 using ImGuiNET;
 using T3.Core;
 using T3.Core.IO;
@@ -51,20 +52,19 @@ namespace T3.Gui.Graph.Interaction
         {
             if (!IsOpen)
             {
-                if (!ImGui.IsWindowFocused() || !ImGui.IsKeyReleased((ImGuiKey)Key.Tab))
+                if (!ImGui.IsWindowFocused() || !ImGui.IsKeyReleased((ImGuiKey)Key.Tab) || !ImGui.IsWindowHovered())
                     return;
 
-                if (NodeSelection.GetSelectedChildUis().Count() == 1)
-                {
-                    var childUi = NodeSelection.GetSelectedChildUis().ToList()[0];
-                    {
-                        var instance = NodeSelection.GetInstanceForSymbolChildUi(childUi);
-                        ConnectionMaker.OpenBrowserWithSingleSelection(this, childUi, instance);
-                    }
-                }
-                else
+                if (NodeSelection.GetSelectedChildUis().Count() != 1)
                 {
                     OpenAt(GraphCanvas.Current.InverseTransformPositionFloat(ImGui.GetIO().MousePos + new Vector2(-4, -20)), null, null, false, null);
+                    return;
+                }
+                
+                var childUi = NodeSelection.GetSelectedChildUis().ToList()[0];
+                {
+                    var instance = NodeSelection.GetInstanceForSymbolChildUi(childUi);
+                    ConnectionMaker.OpenBrowserWithSingleSelection(this, childUi, instance);
                 }
 
                 return;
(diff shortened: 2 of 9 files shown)


Core/Resource/MathUtils.cs
```diff
@@ -1,5 +1,6 @@
 using System;
 using System.Numerics;
+using System.Reflection;
```
domportera, line 3: I dont know how this ended up here and I entirely blame visual studio

Core/Resource/MathUtils.cs
```diff
@@ -146,6 +147,28 @@ public static double RemapAndClamp(double value, double inMin, double inMax, dou
...
+        {
+            int index = startIndex + jumpAmount;
+            index %= collection.Count;
+
+            bool newIndexIsNegative = index < 0;
+            int negativeWrapIndex = realWrap ? collection.Count + index : collection.Count - 1;
+
+            return newIndexIsNegative ? negativeWrapIndex : index;
+        }
+
```
pixtur, line 171: I personally think that this method is still too special to be added to Core.MathUtils. 
Is it used in other places than the SymbolBrowser?

T3/Gui/Graph/Interaction/SymbolBrowser.cs
```diff
@@ -76,62 +76,68 @@ public void Draw()
...
+                var posInWindow = GraphCanvas.Current.ChildPosFromCanvas(PosOnCanvas);
                 _posInScreen = GraphCanvas.Current.TransformPosition(PosOnCanvas);
                 _drawList = ImGui.GetWindowDrawList();
 
                 ImGui.SetNextWindowFocus();
-                DrawMatchesList();
-                DrawSearchInput();
+
+                Vector2 browserPositionInWindow;
+                ShiftPositionToFitOnCanvas(out browserPositionInWindow, posInWindow + _browserPositionOffset, _resultListSize);
```
pixtur, line 86: Nice solution. But why an out param instead of a Vector2 return value?

T3/Gui/Graph/Interaction/SymbolBrowser.cs
```diff
@@ -51,20 +52,19 @@ public void Draw()
         {
             if (!IsOpen)
             {
-                if (!ImGui.IsWindowFocused() || !ImGui.IsKeyReleased((ImGuiKey)Key.Tab))
+                if (!ImGui.IsWindowFocused() || !ImGui.IsKeyReleased((ImGuiKey)Key.Tab) || !ImGui.IsWindowHovered())
```
pixtur, line 55: `|| !ImGui.IsWindowHovered()`

This sounds scary. Did you test this thoroughly for side effects?

T3/Gui/Graph/Interaction/SymbolBrowser.cs
```diff
@@ -287,12 +293,129 @@ private void DrawMatchesList()
...
+
+        /// <summary>
+        /// A method to give some time before expiring a hovered symbol.
+        /// This is intended to allow the cursor to cross over the scroll bar
+        /// of the symbol browser so examples can be interacted with
+        /// </summary>
+        /// <param name="hoveredSymbolUi"></param>
+        async void ExpireLastHoveredSymbolUi(SymbolUi hoveredSymbolUi)
+        {
+            //if we've already started calculating this one, we can just return
```
pixtur, line 327: I know it's nitpicky, but ideally comments should be with space an upper case. 
This helps me to distinguish it from commented code

```suggestion
            // If we've already started calculating this one, we can just return
```

T3/Gui/Graph/Interaction/SymbolBrowser.cs
```diff
@@ -287,12 +293,129 @@ private void DrawMatchesList()
...
+            return symbolUi;
+
+        }
+
+        /// <summary>
+        /// A method to give some time before expiring a hovered symbol.
+        /// This is intended to allow the cursor to cross over the scroll bar
+        /// of the symbol browser so examples can be interacted with
+        /// </summary>
+        /// <param name="hoveredSymbolUi"></param>
```
pixtur, line 324: If there is nothing to say, just remove this line.

T3/Gui/Graph/Interaction/SymbolBrowser.cs
```diff
@@ -287,12 +293,129 @@ private void DrawMatchesList()
...
+        async void ExpireLastHoveredSymbolUi(SymbolUi hoveredSymbolUi)
```
pixtur, line 325: Wow. It took me a while to understand what this method is doing. 
Do we really need such a complex solution?

pixtur: I think you have to feel more comfortable with immediate user interface: The method is called every frame. So it would be MUCH easier to just store the start time and the check on every frame (I.e. on every redraw) if the time has exceeded. i'm pretty sure there is no need for async here.

pixtur (approved):
I just looked at the code here on github.

Besides the definition of the settings and the async method everything looks great.
I added some comments though.

Great work with the symbol browser. But if in doubt, I would prefer smaller PRs.

pixtur:
I did some testing with the SymbolBrowser fixes and found a couple of glitches:

We you want to append another op to an existing op, the layout is broken:

e.g.:

1. Create a [Group]
2. Now press TAB or click on output.

image

Also the Layout of the symbol browser input field should be horizontally aligned with the existing node. The idea behind this is, that the search field is an "operator in the making". If roughly fills the place the new op is going to have.

This is even more noticable when splitting a connection like:

image

I'm also not sure, if flipping the description layout is a good work around. Ideally I would prefer panning the canvas to make space or even not showing the description at all.

T3/Gui/Styling/CustomComponents.cs
```diff
@@ -473,22 +473,33 @@ public static bool DrawSegmentedToggle(ref int currentIndex, List<string> option
...
+
             ImGui.SameLine();
             ImGui.SetCursorPosX(leftPadding + 20);
-            var size = new Vector2(150, ImGui.GetFrameHeight());
+            var modified = DrawSingleValueEdit(cleanedLabel, ref value, min, max, clamp, scale);
+
+            return modified;
+        }
+
+        public static bool FloatValueEditInPlace(string label, ref float value, float min = float.NegativeInfinity, float max = float.PositiveInfinity, float scale = 0.01f, bool clamp = false)
```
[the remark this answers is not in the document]
domportera: agreed i didnt love this and actually this function should have been deleted altogether, my bad!

T3/Gui/UiHelpers/UIControlledSetting.cs
```diff
@@ -0,0 +1,92 @@
...
+            _hiddenUniqueLabel = $"##{label}{_countForUniqueID--}";
+            _uniqueLabel = $"{label}##{_countForUniqueID--}";
+
+            _guiFunc = guiFunc;
+            _tooltip = tooltip;
+            _additionalNotes = additionalNotes;
+            _OnValueChanged = OnValueChanged;
+            DrawOnLeft = drawOnLeft;
+        }
+
```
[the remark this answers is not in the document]
domportera: im interested in your take on this, as to me this seems like a good way to enable Tooll contributors to add/remove user-facing settings with ease. i'm not deeply familiar with the model vs view concept though i think i know what you mean and share the sentiment in most cases. this case to me acts entirely as "view" - just a simple and consistent way to do so. Though with `OnValueChanged` it does allow that scope to creep

I like construction with object initializers too, but I chose to go this route due to the auto-generated unique labeling system (used to prevent certain [imgui bugs](https://github.com/ocornut/imgui/issues/74) that can be easy to run into). You'll notice that when making use of this constructor elsewhere, I essentially treat it like an object initializer, especially due to the optional parameters. Like this:

``
            new UIControlledSetting
            (
                label: "UI Scale",
                tooltip: "The global scale of all rendered UI in the application",
                guiFunc: (string guiLabel) => CustomComponents.DrawSingleValueEdit..... etc
            ),
```

I also want to make sure that anyone creating settings can't mess it up - they know they need at least two things (a label and a draw function), and if they have that it cant go wrong

File: Core/Resource/MathUtils.cs (hunk 2 above)
Comment (domportera) [reply Core/Resource/MathUtils.cs:171]: That's valid. Currently it's not, though i remember seeing another place it could have been. I can move it into the Symbol Browser if you prefer, or as an extension method for IList

File: T3/Gui/Graph/Interaction/SymbolBrowser.cs (hunk 4 above)
Comment (domportera) [reply T3/Gui/Graph/Interaction/SymbolBrowser.cs:55]: I did test it, but i cant say i put it through the wringer. I'm not married to this idea, it just seemed strange that if my mouse is off-screen, i can press Tab and Enter and I can find out 10 minutes later I made an Execute node way off screen

File: T3/Gui/Graph/Interaction/SymbolBrowser.cs (hunk 5 above)
Comment (domportera) [reply T3/Gui/Graph/Interaction/SymbolBrowser.cs:327]: Yes ive been trying to remember this, my bad!

File: T3/Gui/Graph/Interaction/SymbolBrowser.cs (hunk 7 above)
Comment (domportera) [reply T3/Gui/Graph/Interaction/SymbolBrowser.cs:325]: yes it's very complex, I was wrestling with this for a while and I'm also not a huge fan of it. In retrospect this can very much be done in-line or non-async, but I liked the way the async method kept the timing logic and variables in one place. if this was that difficult for you of all people, i think it's definitely worth doing the other way

I can also remove the expiration-timer feature altogether - I implemented this in the case that you didnt want the symbol browser to keep the description open when the symbol browser was up

**Comment by domportera** [comment]
yeah sorry this pull request is so huge - i just ended up getting kind of scattershot. i'm new to this whole open-source-collaboration thing, and i'll be better about the size of my requests in the future

for your last comment, understood! i didnt take that into account but fortunately I baked in an easy way to offset it to align correctly again.

my changes are very much from the perspective of a new user, as I am one myself. personally, I always want to be able to see the descriptions, otherwise I often wont know why they're not there or if they exist at all if it's my first time using the program. But i can make that change no problem

overall it seems like my changes would require a different approach and are at odds with the paradigm here. The idea is for that input field to essentailly be the next node, so in that case its positioning takes priority. so there are a couple of things that can be done to keep the symbol browser usable when at the ends of the screen

At the bottom of the screen, the symbol browser could show above the input field, so it remains in full view while the input field is in the correct position

Or, at the bottom of the screen, the symbol browser could be shorter so that you can scroll, even though your view of the browser would be significantly limited

At the right side of the screen, the symbol browser can shift to stay on screen while the input field remains where it should be

As for the description panel, I can just make it disappear on the right again

File: T3/Gui/Graph/Interaction/SymbolBrowser.cs (hunk 3 above)
Comment (domportera) [reply T3/Gui/Graph/Interaction/SymbolBrowser.cs:86]: just because the description of what it is doing - if it's shifting something, a return value doesn't necessarily suggest that as intuitively as passing in a reference to something that will be shifted

Comment (domportera) [reply T3/Gui/Graph/Interaction/SymbolBrowser.cs:86]: but even by my own description that's not really any clearer and i should just be passing a `ref` if i want to go that route

that function is under construction per your feedback and i'll keep this in mind

**Comment by domportera** [comment]
I adjusted a lot of this to fit your feedback and to fix some bugs I think neither of us were aware of too

to summarize:
- moved the Wrap function to a new class - `CollectionUtils`
- fixed some SymbolBrowser bugs I made 🙈
- made the SymbolBrowser text field be where it's supposed to be
- made the SymbolBrowser shrink vertically at the bottom of the screen so it's still navigable
- made the SymbolBrowser disable the description on default if there is no room, but added a setting to be able to allow it to shift to the left
    - for new users I think this behavior should be enabled by default, because with the default layout the description panel won't show up on almost the entire right half of the graph
-  some minor optimizations in my SymbolBrowser logic
- no more async, much more sane. gotta stop doing this stuff at 4am 😬
- some other readability/style changes you mentioned
- made Settings tooltips work as expected

**Comment by pixtur** [comment]
Awesome, thanks a lot. Much better. I'm merging it on dev.
Some of the settings code I might adjust to a more functional ImGui style, though.

**Comment by pixtur** [comment]
Merged

**Comment by domportera** [comment]
> Some of the settings code I might adjust to a more functional ImGui style, though.

looking forward to seeing your approach!

**Comment by pixtur** [comment]
It's basically a series of function calls. But the code is several hundred lines shorter.

```
            if (ImGui.TreeNode("User Interface"))
            {
                changed |= CustomComponents.DrawCheckboxParameter("Warn before Lib modifications",
                                                       ref UserSettings.Config.WarnBeforeLibEdit,
                                                       "This warning pops up when you attempt to enter an Operator that ships with the application.\n" +
                                                       "If unsure, this is best left checked.");
                
                changed |= CustomComponents.DrawCheckboxParameter("Use arc connections",
                                                                  ref UserSettings.Config.UseArcConnections,
                                                                  "Affects the shape of the connections between your Operators");
                

                changed |= CustomComponents.DrawCheckboxParameter("Use Jog Dial Control",
                                                                  ref UserSettings.Config.UseJogDialControl,
                                                                  "Affects the shape of the connections between your Operators");
```

I also cleaned up some method names in `CustomComponents.cs`.

domportera:
It looks good! yeah I think I got too attached to my other solution with the checkbox-on-right approach, trying to make the settings look good regardless of what order they're in. in lieu of that, this approach is best. Sorry for overcomplicating an otherwise good task to onboard anyone else 😅

pixtur:
Thanks again for the contribution! I believe that one of the greatest benefits of an open source project is the code ping pong. After a couple of iterations the code always gets better, so without your contribution the settings window would not only still be untouched, but there would also be no tooltips and the custom components would still be a mess.

I personally find it really hard to strike the right balance between being too careful (read: "too attached", "scared", "hesitant", "lazy") and too experimental (read: "use stuff without understanding it thoroughly", "change things because it's fun", "recklessness"). Frequently I'm too far on the "careful" side because people will complain when things break. But when too careful, we are stuck and everything fades into being outdated and obsolete.

PR #209, opened 2022-10-31
Fixed standalone version compilation issues

HolgerFoerterer:
* Operators had to be compiled with /unsafe option
* Some dependencies (OpenGL, CppSharp) were missing in the StartT3 project


pixtur:
Thanks!

PR #210, opened 2022-10-31
Added Spout input

HolgerFoerterer:
The texture formats currently known to work are R8G8B8A8_UNorm, B8G8R8A8_UNorm, R16G16B16A16_UNorm and R16G16B16A16_Float. Others may theoretically work but have not been tested.


Operators/Types/lib/io/video/SpoutInput.cs
```diff
@@ -0,0 +1,329 @@
...
+
+        public SpoutInput()
+        {
+            Texture.UpdateAction = Update;
+            _instance++;
+        }
+
+        private void Update(EvaluationContext context)
+        {
+            Command.GetValue(context);
```
pixtur, line 31: I believe this input is not required. The operator will be updated on every frame because of the DirtyFlagTrigger on the output. Could it be, that this is a left over from [SpoutOutput]?

Operators/Types/lib/io/video/SpoutInput.cs
```diff
@@ -0,0 +1,329 @@
...
+                    Texture2D readTexture = new Texture2D(_spoutDX.SenderTexture.__Instance);
+                    immediateContext.CopyResource(readTexture, readableImage);
+                    _currentIndex = (_currentIndex + 1) % NumTextureEntries;
+
+                    Texture.Value = readableImage;
+                }
+            }
+            catch (Exception e)
+            {
+                Log.Debug("Spout input texture reception failed: " + e.ToString());
```
pixtur, line 265:
```suggestion
                Log.Debug("Spout input texture reception failed: " + e.ToString(), SymbolChildId);
```
This would highlight / center the operator when clicking on the message in the log window.

pixtur (changes requested):
Nice!

pixtur (approved):

pixtur:
thank you!

Operators/Types/lib/io/video/SpoutInput.cs
HolgerFoerterer: Oh, yes it is.

PR #243, opened 2023-04-03
Startup optimization

domportera:
Here's the PR for my work with startup optimization. I feel like I've tested it pretty well and it's pretty reliable. I've gotten my personal startup time down from 16-19 seconds to 6.4-6.8 seconds in Release mode.

now there are a LOT of changes, including some unrelated changes regarding consistency or other low-hanging fruit. I commented a lot to help make this more digestible.

I definitely recommend testing this first. It has worked just swell for me but of course you can't be too careful with something like this.

Looking forward to your feedback!

### DIFF
--- a/Core/DataTypes/StructuredList.cs
+++ b/Core/DataTypes/StructuredList.cs
@@ -40,6 +40,14 @@ namespace T3.Core.DataTypes
         public abstract void Write(JsonTextWriter writerNotUsed);
 
         public abstract StructuredList Read(JsonTextReader reader);
+        
+
+        protected static readonly JsonLoadSettings LoadSettings = new()
+                                                                    {
+                                                                        CommentHandling = CommentHandling.Ignore,
+                                                                        DuplicatePropertyNameHandling = DuplicatePropertyNameHandling.Error,
+                                                                        LineInfoHandling = LineInfoHandling.Ignore
+                                                                    };
     }
 
     public class StructuredList<T> : StructuredList where T : struct
@@ -208,7 +216,7 @@ namespace T3.Core.DataTypes
 
         public override StructuredList Read(JsonTextReader reader)
         {
-            var inputToken = JToken.ReadFrom(reader);
+            var inputToken = JToken.ReadFrom(reader, LoadSettings);
 
             var jArray = (JArray)inputToken["StructuredList"];
             if (jArray == null)
--- a/Core/Operator/Instance.cs
+++ b/Core/Operator/Instance.cs
@@ -74,28 +74,28 @@ namespace T3.Core.Operator
 
         public bool AddConnection(Symbol.Connection connection, int multiInputIndex)
         {
-            var (_, sourceSlot, _, targetSlot) = GetInstancesForConnection(connection);
-            if (targetSlot != null)
-            {
-                targetSlot.AddConnection(sourceSlot, multiInputIndex);
-                return true;
-            }
+            var foundInstance = TryGetInstancesForConnection(connection, out _, out var sourceSlot, out _, out var targetSlot);
 
-            return false;
+            if (!foundInstance)
+                return false;
+            
+            targetSlot.AddConnection(sourceSlot, multiInputIndex);
+            return true;
         }
 
         public void RemoveConnection(Symbol.Connection connection, int index)
         {
-            var (_, _, _, targetSlot) = GetInstancesForConnection(connection);
+            var foundInstance = TryGetInstancesForConnection(connection, out _, out _, out _, out var targetSlot);
+            if (!foundInstance)
+                return;
             targetSlot.RemoveConnection(index);
         }
 
-        private (Instance, ISlot, Instance, ISlot) GetInstancesForConnection(Symbol.Connection connection)
+        private bool TryGetInstancesForConnection(Symbol.Connection connection, out Instance sourceInstance, out ISlot sourceSlot, out Instance targetInstance, out ISlot targetSlot)
         {
             Instance compositionInstance = this;
 
-            var sourceInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.SourceParentOrChildId);
-            ISlot sourceSlot;
+            sourceInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.SourceParentOrChildId);
             if (sourceInstance != null)
             {
                 sourceSlot = sourceInstance.Outputs.SingleOrDefault(output => output.Id == connection.SourceSlotId);
@@ -105,14 +105,16 @@ namespace T3.Core.Operator
                 if (connection.SourceParentOrChildId != Guid.Empty)
                 {
                     Log.Error($"connection has incorrect Source: { connection.SourceParentOrChildId}");
-                    return (null, null, null, null);
+                    sourceSlot = null;
+                    targetInstance = null;
+                    targetSlot = null;
+                    return false;
                 }
                 sourceInstance = compositionInstance;
                 sourceSlot = sourceInstance.Inputs.SingleOrDefault(input => input.Id == connection.SourceSlotId);
             }
 
-            var targetInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.TargetParentOrChildId);
-            ISlot targetSlot;
+            targetInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.TargetParentOrChildId);
             if (targetInstance != null)
             {
                 targetSlot = targetInstance.Inputs.SingleOrDefault(e => e.Id == connection.TargetSlotId);
@@ -124,7 +126,7 @@ namespace T3.Core.Operator
                 targetSlot = targetInstance.Outputs.SingleOrDefault(e => e.Id == connection.TargetSlotId);
             }
 
-            return (sourceInstance, sourceSlot, targetInstance, targetSlot);
+            return true;
         }
     }
 
--- a/Core/Operator/Instance.cs
+++ b/Core/Operator/Instance.cs
@@ -72,11 +72,12 @@ namespace T3.Core.Operator
             }
         }
 
-        public bool AddConnection(Symbol.Connection connection, int multiInputIndex)
+        public bool TryAddConnection(Symbol.Connection connection, int multiInputIndex)
         {
-            var foundInstance = TryGetInstancesForConnection(connection, out _, out var sourceSlot, out _, out var targetSlot);
+            var gotSource = TryGetSourceSlot(connection, out var sourceSlot);
+            var gotTarget = TryGetTargetSlot(connection, out var targetSlot);
 
-            if (!foundInstance)
+            if (!gotSource || !gotTarget)
                 return false;
             
             targetSlot.AddConnection(sourceSlot, multiInputIndex);
@@ -85,48 +86,56 @@ namespace T3.Core.Operator
 
         public void RemoveConnection(Symbol.Connection connection, int index)
         {
-            var foundInstance = TryGetInstancesForConnection(connection, out _, out _, out _, out var targetSlot);
-            if (!foundInstance)
+            var success = TryGetTargetSlot(connection, out var targetSlot);
+            if (!success)
                 return;
             targetSlot.RemoveConnection(index);
         }
 
-        private bool TryGetInstancesForConnection(Symbol.Connection connection, out Instance sourceInstance, out ISlot sourceSlot, out Instance targetInstance, out ISlot targetSlot)
+        private bool TryGetSourceSlot(Symbol.Connection connection, out ISlot sourceSlot)
         {
-            Instance compositionInstance = this;
+            var compositionInstance = this;
 
-            sourceInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.SourceParentOrChildId);
-            if (sourceInstance != null)
+            // Get source Instance
+            var sourceInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.SourceParentOrChildId);
+            var gotSourceInstance = sourceInstance != null;
+
+            // Evaluate correctness of slot source Instance
+            var connectionBelongsToThis = connection.SourceParentOrChildId == Guid.Empty;
+            if (!gotSourceInstance && !connectionBelongsToThis)
             {
-                sourceSlot = sourceInstance.Outputs.SingleOrDefault(output => output.Id == connection.SourceSlotId);
-            }
-            else
-            {
-                if (connection.SourceParentOrChildId != Guid.Empty)
-                {
-                    Log.Error($"connection has incorrect Source: { connection.SourceParentOrChildId}");
-                    sourceSlot = null;
-                    targetInstance = null;
-                    targetSlot = null;
-                    return false;
-                }
-                sourceInstance = compositionInstance;
-                sourceSlot = sourceInstance.Inputs.SingleOrDefault(input => input.Id == connection.SourceSlotId);
+                Log.Error($"Connection has incorrect source slot: {connection.SourceParentOrChildId}");
+                sourceSlot = null;
+                return false;
             }
 
-            targetInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.TargetParentOrChildId);
-            if (targetInstance != null)
-            {
-                targetSlot = targetInstance.Inputs.SingleOrDefault(e => e.Id == connection.TargetSlotId);
-            }
-            else
+            // Get source Slot
+            var sourceSlotList = gotSourceInstance ? sourceInstance.Outputs : compositionInstance.Inputs.Cast<ISlot>();
+            sourceSlot = sourceSlotList.SingleOrDefault(slot => slot.Id == connection.SourceSlotId);
+            return sourceSlot is not null;
+        }
+
+        private bool TryGetTargetSlot(Symbol.Connection connection, out ISlot targetSlot)
+        {
+            var compositionInstance = this;
+              
+            // Get target Instance
+            var targetInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.TargetParentOrChildId);
+            var gotTargetInstance = targetInstance is not null;
+
+            // Get target Slot
+            var targetSlotList = gotTargetInstance ? targetInstance.Inputs.Cast<ISlot>() : compositionInstance.Outputs;
+            targetSlot = targetSlotList.SingleOrDefault(slot => slot.Id == connection.TargetSlotId);
+            var gotTargetSlot = targetSlot is not null;
+            
+            #if DEBUG
+            if (!gotTargetInstance)
             {
                 Debug.Assert(connection.TargetParentOrChildId == Guid.Empty);
-                targetInstance = compositionInstance;
-                targetSlot = targetInstance.Outputs.SingleOrDefault(e => e.Id == connection.TargetSlotId);
             }
-
-            return true;
+            #endif
+            
+            return gotTargetSlot;
         }
     }
 
--- a/Core/Operator/Slots/InputValue.cs
+++ b/Core/Operator/Slots/InputValue.cs
@@ -86,7 +86,7 @@ namespace T3.Core.Operator.Slots
             }
             else
             {
-                //Log.Error($"Trying to read a json value for type '{ValueType}' but no converter registered in JsonToTypeValueConverters. Skipping value reading.");
+                //T3.Core.Logging.Log.Error($"Trying to read a json value for type '{ValueType}' but no converter registered in JsonToTypeValueConverters. Skipping value reading.");
             }
         }
 
--- a/Core/Operator/Slots/Slot.cs
+++ b/Core/Operator/Slots/Slot.cs
@@ -1,5 +1,6 @@
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using T3.Core.DataTypes;
 using T3.Core.Logging;
 using T3.Core.Stats;
@@ -93,16 +94,15 @@ namespace T3.Core.Operator.Slots
 
         public void AddConnection(ISlot sourceSlot, int index = 0)
         {
-            if (!IsConnected && sourceSlot != null)
+            Debug.Assert(sourceSlot is not null);
+            
+            if (!IsConnected)
             {
                 _actionBeforeAddingConnecting = UpdateAction;
                 UpdateAction = ConnectedUpdate;
                 DirtyFlag.Target = sourceSlot.DirtyFlag.Target;
                 DirtyFlag.Reference = DirtyFlag.Target - 1;
             }
-
-            if (sourceSlot == null)
-                return;
             
             if (sourceSlot.ValueType != ValueType)
             {
--- a/Core/Operator/Symbol.cs
+++ b/Core/Operator/Symbol.cs
@@ -1,8 +1,10 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using T3.Core.Logging;
 using T3.Core.Operator.Attributes;
 using T3.Core.Operator.Slots;
@@ -433,6 +435,7 @@ namespace T3.Core.Operator
         {
             // order the inputs by the given input definitions. original order is coming from code, but input def order is the relevant one
             int numInputs = inputs.Count;
+            
             for (int i = 0; i < numInputs - 1; i++)
             {
                 Guid inputId = InputDefinitions[i].Id;
@@ -460,49 +463,45 @@ namespace T3.Core.Operator
 
             SortInputSlotsByDefinitionOrder(newInstance.Inputs);
 
-            // create child instances
-            foreach (var symbolChild in Children)
+            foreach (var child in Children)
             {
-                CreateAndAddNewChildInstance(symbolChild, newInstance);
+                CreateAndAddNewChildInstance(child, newInstance);
             }
 
             // create connections between instances
-            var conHashToCount = new Dictionary<ulong, int>(Connections.Count);
-            var invalidConnections = new List<Connection>();
-            foreach (var connection in Connections)
+            if (Connections.Count != 0)
             {
-                ulong highPart = 0xFFFFFFFF & (ulong)connection.TargetSlotId.GetHashCode();
-                ulong lowPart = 0xFFFFFFFF & (ulong)connection.TargetParentOrChildId.GetHashCode();
-                ulong hash = (highPart << 32) | lowPart;
-                if (!conHashToCount.TryGetValue(hash, out int count))
-                    conHashToCount.Add(hash, 0);
-
-                var valid = newInstance.AddConnection(connection, count);
-                if (!valid)
+                var conHashToCount = new Dictionary<ulong, int>(Connections.Count);
+                for (var index = 0; index < Connections.Count; index++) // warning: the order in which these are processed matters
                 {
-                    Log.Warning("Skipping connection to no longer existing targets");
-                    invalidConnections.Add(connection);
-                    continue;
-                }
- 
-                conHashToCount[hash] = count + 1;
-            }
+                    var connection = Connections[index];
+                    ulong highPart = 0xFFFFFFFF & (ulong)connection.TargetSlotId.GetHashCode();
+                    ulong lowPart = 0xFFFFFFFF & (ulong)connection.TargetParentOrChildId.GetHashCode();
+                    ulong hash = (highPart << 32) | lowPart;
+                    if (!conHashToCount.TryGetValue(hash, out int count))
+                        conHashToCount.Add(hash, 0);
 
-            foreach (var c in invalidConnections)
-            {
-                Connections.Remove(c);
+                    var valid = newInstance.AddConnection(connection, count);
+                    if (!valid)
+                    {
+                        Log.Warning("Skipping connection to no longer existing targets");
+                        Connections.RemoveAt(index);
+                        index--;
+                        continue;
+                    }
+
+                    conHashToCount[hash] = count + 1;
+                }
             }
 
             // connect animations if available
             Animator.CreateUpdateActionsForExistingCurves(newInstance.Children);
-
             InstancesOfSymbol.Add(newInstance);
 
             return newInstance;
         }
 
-
-        private Instance CreateAndAddNewChildInstance(SymbolChild symbolChild, Instance parentInstance)
+        private static Instance CreateAndAddNewChildInstance(SymbolChild symbolChild, Instance parentInstance)
         {
             var childSymbol = symbolChild.Symbol;
             var childInstance = childSymbol.CreateInstance(symbolChild.Id);
@@ -513,27 +512,30 @@ namespace T3.Core.Operator
             {
                 Debug.Assert(i < childInstance.Inputs.Count);
                 Guid inputDefinitionId = childSymbol.InputDefinitions[i].Id;
-                childInstance.Inputs[i].Input = symbolChild.InputValues[inputDefinitionId];
-                childInstance.Inputs[i].Id = inputDefinitionId;
+                var inputSlot = childInstance.Inputs[i];
+                inputSlot.Input = symbolChild.InputValues[inputDefinitionId];
+                inputSlot.Id = inputDefinitionId;
             }
 
             // set up the outputs for the child instance
             for (int i = 0; i < symbolChild.Symbol.OutputDefinitions.Count; i++)
             {
                 Debug.Assert(i < childInstance.Outputs.Count);
-                childInstance.Outputs[i].Id = childSymbol.OutputDefinitions[i].Id;
-                var symbolChildOutput = symbolChild.Outputs[childInstance.Outputs[i].Id];
-                if (childSymbol.OutputDefinitions[i].OutputDataType != null)
+                var outputSlot = childInstance.Outputs[i];
+                var outputDefinition = childSymbol.OutputDefinitions[i];
+                outputSlot.Id = outputDefinition.Id;
+                var symbolChildOutput = symbolChild.Outputs[outputSlot.Id];
+                if (outputDefinition.OutputDataType != null)
                 {
                     // output is using data, so link it
-                    if (childInstance.Outputs[i] is IOutputDataUser outputDataConsumer)
+                    if (outputSlot is IOutputDataUser outputDataConsumer)
                     {
                         outputDataConsumer.SetOutputData(symbolChildOutput.OutputData);
                     }
                 }
 
-                childInstance.Outputs[i].DirtyFlag.Trigger = symbolChildOutput.DirtyFlagTrigger;
-                childInstance.Outputs[i].IsDisabled = symbolChildOutput.IsDisabled;
+                outputSlot.DirtyFlag.Trigger = symbolChildOutput.DirtyFlagTrigger;
+                outputSlot.IsDisabled = symbolChildOutput.IsDisabled;
             }
 
             parentInstance.Children.Add(childInstance);
--- a/Core/Operator/Symbol.cs
+++ b/Core/Operator/Symbol.cs
@@ -1,10 +1,8 @@
 using System;
-using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
-using System.Threading;
 using T3.Core.Logging;
 using T3.Core.Operator.Attributes;
 using T3.Core.Operator.Slots;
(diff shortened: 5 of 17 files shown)


Core/DataTypes/StructuredList.cs
domportera, line 1: This can be largely ignored, I just ran into it somehow while searching around. Just trying to make `JsonLoadSettings` a thing so there is no ambiguity.

Core/Operator/Slots/InputValue.cs
domportera, line 1: Also can be ignored, self explanatory but I used it during my debugging

Core/Operator/Slots/Slot.cs
domportera, line 1: In my travels I discovered that there is no circumstance where this function is called and `sourceSlot` is null, so I removed the checks (except in Debug mode).

Core/Resource/JsonResult.cs
domportera, line 1: In order to properly parallelize things, I created a generic class that handles reading out a `JToken` from a provided file. This provides a very easy way to separate and parallelize file reading, and ideally we could do the same for file writing as well.

When json is loaded elsewhere in the project this could be handy too.

Core/Resource/ResourceManager.cs
```diff
@@ -466,7 +466,7 @@ public uint CreateVertexShaderFromFile(string srcFile, string entryPoint, string
 
         public bool CreatePixelShaderFromSource(string shaderSource, string name, string entryPoint, ref uint resourceId)
         {
-            if (string.IsNullOrEmpty(shaderSource) || string.IsNullOrEmpty(entryPoint))
+            if (string.IsNullOrWhiteSpace(shaderSource) || string.IsNullOrWhiteSpace(entryPoint))
```
domportera, line 469: I ran into this while testing - if you have a [CustomPixelShader] and backspace everything and just press Space, T3 crashes. Now it doesn't :)

Editor/Gui/Graph/GraphCanvas.cs
```diff
@@ -969,15 +969,11 @@ private void CopySelectedNodesToClipboard()
...
-
-                var jsonUi = new SymbolUiJson { Writer = json.Writer };
-                jsonUi.WriteSymbolUi(newContainerUi);
-
-                json.Writer.WriteEndArray();
+                var jsonWriter = new JsonTextWriter(writer);
+                jsonWriter.WriteStartArray();
+                SymbolJson.WriteSymbol(containerOp, jsonWriter);
+                SymbolUiJson.WriteSymbolUi(newContainerUi, jsonWriter);
+                jsonWriter.WriteEndArray();
```
domportera, line 976: update to use new static SymbolUiJson functions, also cleaner

Editor/Gui/Graph/GraphCanvas.cs
domportera, line 1: update to use new static SymbolUiJson functions, also cleaner

Editor/Gui/Graph/GraphCanvas.cs
```diff
@@ -1038,29 +1047,29 @@ private void PasteClipboard()
...
         private void DrawGrid()
         {
-            var color = new Color(0, 0, 0, 0.15f);
             var gridSize = Math.Abs(64.0f * Scale.X);
             for (var x = (-Scroll.X * Scale.X) % gridSize; x < WindowSize.X; x += gridSize)
             {
                 DrawList.AddLine(new Vector2(x, 0.0f) + WindowPos,
                                  new Vector2(x, WindowSize.Y) + WindowPos,
-                                 color);
+                                 GridColor);
```
domportera, line 1064: Stupid little thing that popped out at me when I was working on what's above.

a little more readable and the micro-est of optimizations

Editor/Gui/Graph/Interaction/ExampleSymbolLinking.cs
domportera, line 1: Sped up updating the example links by reducing iteration count

Editor/Gui/Graph/PlayerExporter.cs
domportera, line 1: update to new static SymbolJson

Player/Program.cs
domportera, line 1: by default, I made all of the startup logs not log, controlled by this bool. Warnings and errors still log.

Core/Operator/Instance.cs
domportera, line 1: Optimized instance creation - removing unnecessary null checks and replacing them with `bool` checks where possible.

Core/Operator/Instance.cs
```diff
@@ -74,28 +74,28 @@ where typeof(IInputSlot).IsAssignableFrom(field.FieldType)
...
         {
-            var (_, _, _, targetSlot) = GetInstancesForConnection(connection);
+            var foundInstance = TryGetInstancesForConnection(connection, out _, out _, out _, out var targetSlot);
+            if (!foundInstance)
+                return;
             targetSlot.RemoveConnection(index);
         }
 
-        private (Instance, ISlot, Instance, ISlot) GetInstancesForConnection(Symbol.Connection connection)
+        private bool TryGetInstancesForConnection(Symbol.Connection connection, out Instance sourceInstance, out ISlot sourceSlot, out Instance targetInstance, out ISlot targetSlot)
```
domportera, line 94: a lot of `out`s, but avoids having to check for nulls or create anonymous objects. We simply return `false` if it failed, and `true` if it succeeded.

Core/Operator/Instance.cs
```diff
@@ -74,28 +74,28 @@ where typeof(IInputSlot).IsAssignableFrom(field.FieldType)
...
         {
             Instance compositionInstance = this;
 
-            var sourceInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.SourceParentOrChildId);
-            ISlot sourceSlot;
+            sourceInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.SourceParentOrChildId);
```
domportera, line 98: opportunity for optimization - using `FirstOrDefault`. I did this with the `Symbol` and `SymbolUi` loading stuff, but thought these runtime checks might be necessary so I left them in. The only time I can see this ever needing to check for "Single" is at compile time - like if a dev were to make an operator from code and forget to make a guid unique

Core/Operator/Symbol.cs
```diff
@@ -460,49 +463,45 @@ public Instance CreateInstance(Guid id)
...
+                        Connections.RemoveAt(index);
+                        index--;
+                        continue;
+                    }
 
-            foreach (var c in invalidConnections)
-            {
-                Connections.Remove(c);
+                    conHashToCount[hash] = count + 1;
+                }
```
domportera, line 494: potentially removed several iterations over the list by removing connections from within the loop itself.

fun fact: this is where I broke everything :) took me way too long to realize... hence the comment lmao

Core/Operator/Symbol.cs
```diff
@@ -513,27 +512,30 @@ private Instance CreateAndAddNewChildInstance(SymbolChild symbolChild, Instance
...
             {
                 Debug.Assert(i < childInstance.Outputs.Count);
-                childInstance.Outputs[i].Id = childSymbol.OutputDefinitions[i].Id;
-                var symbolChildOutput = symbolChild.Outputs[childInstance.Outputs[i].Id];
-                if (childSymbol.OutputDefinitions[i].OutputDataType != null)
+                var outputSlot = childInstance.Outputs[i];
+                var outputDefinition = childSymbol.OutputDefinitions[i];
+                outputSlot.Id = outputDefinition.Id;
+                var symbolChildOutput = symbolChild.Outputs[outputSlot.Id];
+                if (outputDefinition.OutputDataType != null)
```
domportera, line 528: reduce unnecessary array indexing

Core/Operator/Symbol.cs
```diff
@@ -1,8 +1,10 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
```
domportera, line 7: oops... these `using`s are leftovers from me screwing around

Core/Resource/Model.cs
```diff
@@ -442,51 +444,86 @@ private static void RegisterType(Type type, string typeName, Func<InputValue> de
...
+                {
+                    Log.Error($"Null or empty namespace of type {newType.Name}");
+                    continue;
+                } 
+                
+                var @namespace = _innerNamespace.Replace(typeNamespace, string.Empty).ToLowerInvariant();
+                var idFromNamespace = _idFromNamespace.Match(typeNamespace).Value;
+                idFromNamespace = ReplaceCharacter(idFromNamespace, '_', '-');
+                
+                Debug.Assert(!string.IsNullOrWhiteSpace(idFromNamespace));
```
domportera, line 511: I optimized this to reduce `string` object allocations as much as possible using the `ReplaceCharacter` function.

Core/Resource/Model.cs
```diff
@@ -495,49 +532,52 @@ where type.IsSubclassOf(typeof(Instance))
...
-
+        
         public virtual void SaveAll()
         {
             ResourceFileWatcher.DisableOperatorFileWatcher(); // don't update ops if file is written during save
-
-            RemoveAllSymbolFiles();
+            
+            // todo: this sounds like a dangerous step. we should overwrite these files by default and can check which files are not overwritten to delete others?
+            RemoveAllSymbolFiles();
```
domportera, line 580: I didn't change this, but one day could be worth doing. it's scary 😨

Core/Resource/Model.cs
```diff
@@ -495,49 +532,52 @@ where type.IsSubclassOf(typeof(Instance))
...
-                return symbol;
+                var c = idReadOnlySpan[i];
+                namespaceIdSpan[i] = c == toReplace ? toReplaceWith : c;
             }
-            catch (Newtonsoft.Json.JsonReaderException e)
+
+            return new string(namespaceIdSpan);
+        }
+        
+        internal bool TryReadSymbolWithId(Guid symbolId, out Symbol symbol)
```
domportera, line 551: A re-do on the `ReadSymbolWithId function using the `JsonResult`s that we already got earlier

Core/Resource/Model.cs
```diff
@@ -638,7 +677,7 @@ private static void RemoveDeprecatedSymbolFiles(Symbol symbol)
             if (string.IsNullOrEmpty(symbol.DeprecatedSourcePath))
                 return;
 
-            foreach (var fileExtension in _operatorFileExtensions)
+            foreach (var fileExtension in OperatorFileExtensions)
```
domportera, line 680: style consistency

Core/Resource/Model.cs
```diff
@@ -710,13 +749,14 @@ public static string BuildFilepathForSymbol(Symbol symbol, string extension)
...
         private static OpUpdateCounter _updateCounter;
-        
+
         public const string SourceExtension = ".cs";
         private const string SymbolExtension = ".t3";
         protected const string SymbolUiExtension = ".t3ui";
         public const string OperatorTypesFolder = @"Operators\Types\";
 
-        private static readonly List<string> _operatorFileExtensions = new()
+        private static readonly Type InstanceType = typeof(Instance);
```
domportera, line 758: remove unnecessary runtime `typeof` calls

Core/Resource/SymbolJson.cs
domportera, line 1: This is now a static class. The only actual members it had were `Writer` and `Reader` - I thought it would be safer, easier, more efficient, and less confusing to just make this a static class. A lot has changed here with no functionality changes as far as I can tell.

Core/Resource/SymbolJson.cs
```diff
@@ -1,10 +1,10 @@
 ﻿using System;
+using System.Collections.Concurrent;
```
domportera, line 2: this might be an oopsie...

Core/Resource/SymbolJson.cs
```diff
@@ -30,218 +30,215 @@ public static void WriteObject(this JsonTextWriter writer, string name, object v
...
-        public void WriteSymbol(Symbol symbol)
+        public static void WriteSymbol(Symbol symbol, JsonTextWriter writer)
         {
-            Writer.WriteStartObject();
+            writer.WriteStartObject();
 
-            Writer.WriteObject("Name", symbol.Name);
-            Writer.WriteValue("Id", symbol.Id);
-            Writer.WriteObject("Namespace", symbol.Namespace);
+            writer.WriteObject(JsonKeys.Name, symbol.Name);
```
domportera, line 41: you'll see this JsonKeys.[whatever] everywhere in this class. It's basically paranoia but I think it makes working with Json in the future a little easier.

The idea is that 
1. now the compiler HAS TO make these keys constants, so it's basically GC-free
2. it's far more difficult to make a typo
3. it provides an easy high-level overview of the keys present in the json we are concerned with

Core/Resource/SymbolJson.cs
```diff
@@ -30,218 +30,215 @@ public static void WriteObject(this JsonTextWriter writer, string name, object v
...
             }
 
-            Writer.WriteEndArray();
+            writer.WriteEndArray();
         }
         #endregion
 
         #region reading
-        private static SymbolChild ReadSymbolChild(Model model, JToken symbolChildJson)
+        private static bool TryReadSymbolChild(Model model, JToken symbolChildJson, out SymbolChild symbolChild)
```
domportera, line 167: Like many other functions so far, I've also changed this to the `TryX` paradigm that has become popular in C#. It vastly reduces the number of null checks we have to perform on the results, and is straightforward about the fact that this function can fail.

Core/Resource/SymbolJson.cs
```diff
@@ -30,218 +30,215 @@ public static void WriteObject(this JsonTextWriter writer, string name, object v
...
-            {
-                Log.Warning($"Failed to load symbol {symbolId}.");
-                return null;
+                var loaded = model.TryReadSymbolWithId(symbolId, out symbol);
+                if (!loaded)
+                {
+                    Log.Warning($"Error loading symbol child {symbolId}");
+                    symbolChild = null;
+                    return false;
+                }
```
domportera, line 180: A single clear point in the function of total failure. If we get past here, we have a valid `Symbol` object.

Core/Resource/SymbolJson.cs
```diff
@@ -255,105 +252,76 @@ private static void ReadChildInputValue(SymbolChild symbolChild, JToken inputJso
...
         {
-            if (json["Type"] != null)
+            if (json[JsonKeys.Type] != null)
             {
                 symbolChild.Outputs[outputId].OutputData.ReadFromJson(json);
             }
         }
 
-        public Symbol ReadSymbol(Model model)
+        public static bool TryReadSymbol(Model model, JToken jToken, out Symbol symbol, bool allowNonOperatorInstanceType)
```
domportera, line 261: the public facing read - currently only used by the `Graph` class I think for pasting operators.

Core/Resource/SymbolJson.cs
```diff
@@ -255,105 +252,76 @@ private static void ReadChildInputValue(SymbolChild symbolChild, JToken inputJso
...
-                        missingSymbolChildIds.Add(childId);
-
-                    if (symbolId != Guid.Empty)
-                        missingSymbolsIds.Add(symbolId);
-                }
-                else
-                {
-                    symbolChildren.Add(symbolChild);
-                }
+                ObtainSymbolChild(model, childJson, symbolChildren, name, missingSymbolChildIds);
```
domportera, line 286: separated to a separate function for readability

Core/Resource/SymbolJson.cs
```diff
@@ -255,105 +252,76 @@ private static void ReadChildInputValue(SymbolChild symbolChild, JToken inputJso
...
-                         || missingSymbolChildIds.Contains(connection.SourceParentOrChildId)
-                    )
-                {
-                    Log.Warning("Skipping connection to child of undefined type");
-                }
-                else
-                {
-                    connections.Add(connection);
-                }
+                ReadConnections(connectionsJson, name, missingSymbolChildIds, connections);
```
domportera, line 294: separated to separate function for readability

Core/Resource/SymbolJson.cs
```diff
@@ -255,105 +252,76 @@ private static void ReadChildInputValue(SymbolChild symbolChild, JToken inputJso
...
-                {
-                    instanceType = typeof(object);
-                }
-                else
-                {
-                    MessageBox.Show($"Definition '{instanceTypeName}' is missing in Operator.dll.\nPlease try to rebuild your solution.");
-                    Application.Exit();
-                    Application.ExitThread();
-                }
+                inputDefaultValues[idAndValue.Item1] = idAndValue.Item2;
```
domportera, line 307: re-ordered some of the work here to remove redundant iterations

Core/Resource/SymbolJson.cs
```diff
@@ -368,9 +336,144 @@ public Symbol ReadSymbol(Model model, JToken o, bool allowNonOperatorInstanceTyp
...
+            public const string OutputData = "OutputData";
+            public const string Type = "Type";
+            public const string DirtyFlagTrigger = "DirtyFlagTrigger";
+            public const string IsDisabled = "IsDisabled";
+            public const string DefaultValue = "DefaultValue";
+            public const string Value = "Value";
+            public const string Inputs = "Inputs";
+            public const string Outputs = "Outputs";
+            public const string Animator = "Animator";
+        }
```
domportera, line 470: Might not be the prettiest way to have a whole bunch of constant strings, but I couldn't think of a prettier way

Core/Resource/SymbolJson.cs
```diff
@@ -368,9 +336,144 @@ public Symbol ReadSymbol(Model model, JToken o, bool allowNonOperatorInstanceTyp
...
+
+        public static JsonLoadSettings LoadSettings { get; } = new()
+                                                                   {
+                                                                       CommentHandling = CommentHandling.Ignore,
+                                                                       DuplicatePropertyNameHandling = DuplicatePropertyNameHandling.Error,
+                                                                       LineInfoHandling = LineInfoHandling.Ignore
+                                                                   };
```
domportera, line 477: consistent rules for loading Symbols

Editor/Gui/UiSymbolJson.cs
domportera, line 1: Basically everything I did for `SymbolJson` I did for this class as well.

Editor/Gui/UiModel.cs
domportera, line 1: Basically everything I did for `Model` I did for this class as well. This class is just simpler because it doesn't have to worry about recursive lookups like the base class does.

Editor/Gui/UiSymbolJson.cs
```diff
@@ -321,33 +315,58 @@ public static SymbolUi ReadSymbolUi(JToken mainObject)
...
                                   };
-            return newSymbolUi;
+            return true;
+        }
+
+        private static readonly Type Vector2Type = typeof(Vector2);
+        private static readonly Type Vector4Type = typeof(Vector4);
+        private static readonly Type StylesType = typeof(SymbolChildUi.Styles);
+        private static readonly Type ConnectionStylesType = typeof(SymbolChildUi.ConnectionStyles);
+
```
domportera, line 352: Cached important type info instead of unnecessarily doing it at runtime

Core/Resource/SymbolJson.cs
```diff
@@ -368,9 +336,144 @@ public Symbol ReadSymbol(Model model, JToken o, bool allowNonOperatorInstanceTyp
...
+                    continue;
+                }
+
+                connections.Add(connection);
+            }
+        }
+
+        private static readonly Type ObjectType = typeof(object);
+        // Define a static dictionary to cache Type instances by instanceTypeName
+        private static readonly ConcurrentDictionary<string, Type> TypeCache = new();
```
domportera, line 445: cached types to remove redundant `typeof` invokations

Core/Operator/Symbol.cs
pixtur: Okay

Core/Operator/Slots/Slot.cs
```diff
@@ -93,16 +94,15 @@ public T GetValue(EvaluationContext context)
 
         public void AddConnection(ISlot sourceSlot, int index = 0)
         {
-            if (!IsConnected && sourceSlot != null)
+            Debug.Assert(sourceSlot is not null);
```
pixtur, line 97: Interesting! Can you make this assumption? I didn't check the code...

Core/Operator/Instance.cs
pixtur: Ah, so you would prefer `FirstOfDefault`? Good question...
I believe this would be premature. A better and much more substantial optimization would be to turn some of theses lists (esp. Symbol.Children) into ordered dictionaries.

Core/Operator/Instance.cs
pixtur: I like that.

Core/Operator/Instance.cs
```diff
@@ -74,28 +74,28 @@ where typeof(IInputSlot).IsAssignableFrom(field.FieldType)
 
         public bool AddConnection(Symbol.Connection connection, int multiInputIndex)
         {
-            var (_, sourceSlot, _, targetSlot) = GetInstancesForConnection(connection);
-            if (targetSlot != null)
-            {
-                targetSlot.AddConnection(sourceSlot, multiInputIndex);
-                return true;
-            }
+            var foundInstance = TryGetInstancesForConnection(connection, out _, out var sourceSlot, out _, out var targetSlot);
```
pixtur, line 77: This code looks so weird... 
I can't help but to think there must be a better way.

Core/DataTypes/StructuredList.cs
```diff
@@ -40,6 +40,14 @@ public StructuredList(Type type)
         public abstract void Write(JsonTextWriter writerNotUsed);
 
         public abstract StructuredList Read(JsonTextReader reader);
+        
+
+        protected static readonly JsonLoadSettings LoadSettings = new()
```
pixtur, line 45: I didn't stumble of the definition of JsonLoadSettings. Does this come from NewtonSoft?

pixtur (changes requested):
Excellent work! I agree with 90% of the changes. 

But some of the performance optimizations appear "premature". I would be really curious to see the actual impact of the `typeof` caching. I'm probably learning something here, but I need some convincing numbers.

pixtur (changes requested):
Excellent work! I believe 90% of the changes are spot on.

But some the optimizations appear premature. Maybe I'm learning something here, but especially for the `typeof(T)` caching I am curious to see some measurements to change my mind.

Core/Resource/SymbolJson.cs
domportera: I'm a big fan of the nameof() idea - it'd definitely be less efficient, but at a future point we can even manage saving by reflection using that way (if that's ever actually necessary or helpful - doubt it). Though it would definitely invalidate our current files if we don't find the right cocktail for ensuring we're excluding class and namespace names. I don't think that's a tall order, though.

On your point about assumption 1, I'm just referencing the fact that string literals are not guaranteed to be compile-time constants - the compiler(s) make those decisions but that's obv a big black box to me. In the event that the compiler decides to allocate new strings for every function call instead of making them constants, that's where the GC could kick in

Core/Resource/SymbolJson.cs
domportera: LOL I think I'm with you in that I feel like it shouldn't be necessary - we should just be able to "copy/paste" references to pointers. Problem for another day I suppose

Core/Resource/SymbolJson.cs
domportera: probably NOT, good point

Core/Resource/SymbolJson.cs
domportera: I just know when I was navigating this, what was default/handled wasn't clear to me. I managed to have comments read out into my string params and so I did this solely for clarity

Editor/Gui/UiSymbolJson.cs
domportera: I did measure this, but I can't tell you exactly what the result was. I basically just A-B'd the total starting time and I remember it being more significant than I even expected. Doing a quick "benchmark" online showed an 8x speedup in type comparison operations, though I wouldn't trust that entirely lmfao

Core/Resource/SymbolJson.cs
domportera: feedback definitely heard, but I imagine the compilers make decisions about compile-time caching of values. I saw a noticeable performance boost from caching some types, I don't remember which, when this was a single-threaded operation. That being said, it's possible that this type caching is no longer beneficial and it should be benchmarked.

Editor/Gui/UiSymbolJson.cs
```diff
@@ -19,216 +19,192 @@
...
         {
-            Writer.WritePropertyName("InputUis");
-            Writer.WriteStartArray();
+            writer.WritePropertyName(JsonKeys.InputUis);
+            writer.WriteStartArray();
 
             foreach (var inputEntry in symbolUi.InputUis)
             {
-                var symbolInput = symbolUi.Symbol.InputDefinitions.SingleOrDefault(inputDef => inputDef.Id == inputEntry.Key);
+                var symbolInput = symbolUi.Symbol.InputDefinitions.FirstOrDefault(inputDef => inputDef.Id == inputEntry.Key);
```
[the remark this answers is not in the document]
domportera: I agree in the principle of failing often with clear errors, but I personally would prefer this to happen on the root-most level, such that the error propogates so you don't have to go searching for its origin and consuming classes don't need to constantly check for valid data.

With json, this stance becomes tricker, since anyone can just open the file and ruin it. then it becomes more of a moral question... 😈

Editor/Gui/UiSymbolJson.cs
```diff
@@ -19,216 +19,192 @@
...
+        private static void WriteAnnotations(SymbolUi symbolUi, JsonTextWriter writer)
         {
             if (symbolUi.Annotations.Count == 0)
                 return;
 
-            var vec2Writer = TypeValueToJsonConverters.Entries[typeof(Vector2)];
-            var vec4Writer = TypeValueToJsonConverters.Entries[typeof(Vector4)];
-            Writer.WritePropertyName("Annotations");
-            Writer.WriteStartArray();
+            var vec2Writer = TypeValueToJsonConverters.Entries[Vector2Type];
```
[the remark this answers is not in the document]
domportera: This overall falls into the type caching convo. That being said, I feel like these entries (and other conversion/ "registered" entries) could be static and computed via an initializer, then maybe accessed as an enum?

Editor/Gui/UiSymbolJson.cs
```diff
@@ -19,216 +19,192 @@
...
+                writer.WriteEndObject();
             }
 
-            Writer.WriteEndArray();
+            writer.WriteEndArray();
         }
-
-        public SymbolUi ReadSymbolUi(string filePath)
+        
+        internal static bool TryReadSymbolUi(JToken mainObject, out SymbolUi symbolUi)
```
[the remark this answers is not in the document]
domportera: SAME
but what i like even more is being so certain as to not needing them :)

Editor/Gui/Graph/GraphCanvas.cs
domportera: valid point, I can un-cache. I just have difficulty seeing the risk, and see the potential for exposing this as a setting at a later date

Core/Operator/Instance.cs
domportera: I'd definitely be a bigger fan of having these be ordered dictionaries too

Editor/Gui/UiSymbolJson.cs
```diff
@@ -293,23 +278,32 @@ public static SymbolUi ReadSymbolUi(JToken mainObject)
...
+                                $"Assuming that the output was removed and ignoring the ui information.");
                     continue;
                 }
 
                 var type = outputDefinition.ValueType;
                 if (OutputUiFactory.Entries.TryGetValue(type, out var outputCreator))
                 {
                     var outputUi = outputCreator();
-                    outputUi.OutputDefinition = symbol.OutputDefinitions.Single(def => def.Id == outputId);
+                    outputUi.OutputDefinition = symbol.OutputDefinitions.First(def => def.Id == outputId);
```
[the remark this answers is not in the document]
domportera: maybe this is the place. id want this to fail but only in specific places. the OrderedDictionary idea probably just solves this in the best way for everyone

Editor/Gui/UiModel.cs
```diff
@@ -242,34 +244,41 @@ private void Init()
...
             // first load core data
-            base.Load();
-
-            var uiJson = new SymbolUiJson();
+            base.Load(log);
 
             var symbolUiFiles = Directory.GetFiles(OperatorTypesFolder, $"*{SymbolUiExtension}", SearchOption.AllDirectories);
-
-            foreach (var symbolUiFile in symbolUiFiles)
+            var symbolUiJsons = symbolUiFiles.AsParallel()
```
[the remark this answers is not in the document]
domportera: I'm not a huge fan either but proper spacing allows things to be readable. I mainly used it because it made "smart" parallelization a breeze

the GC hit happens if you abuse them in hot paths instead of properly iterating. since this is happening on startup, I'm not worried.

though I could name a place function or two where i certainly ignored that advice 😅

Editor/Gui/Graph/Interaction/ExampleSymbolLinking.cs
```diff
@@ -10,15 +11,37 @@ public static class ExampleSymbolLinking
         public static void UpdateExampleLinks()
         {
             ExampleSymbols.Clear();
-            foreach (var symbolUi in SymbolUiRegistry.Entries.Values)
+
+            var symbolCollection = SymbolUiRegistry.Entries.Values;
```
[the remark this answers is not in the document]
domportera: agreed! this was a silly oversight

Core/Resource/SymbolJson.cs
domportera: in the new parallel workflow I'm no longer seeing as large of a performance boost for all the type caching, probably about ~100ms, so I am removing the type cache

Editor/Gui/Graph/GraphCanvas.cs
```diff
@@ -1104,6 +1113,7 @@ public IEnumerable<ISelectableCanvasObject> SelectableChildren
         public static readonly LibWarningDialog LibWarningDialog = new();
 
         //public override SelectionHandler SelectionHandler { get; } = new SelectionHandler();
+        private static readonly Color GridColor = new(0, 0, 0, 0.15f);
```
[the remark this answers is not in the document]
domportera: agreed!

Core/Resource/Model.cs
```diff
@@ -495,49 +532,52 @@ where type.IsSubclassOf(typeof(Instance))
...
         }
-
+        
         public virtual void SaveAll()
         {
             ResourceFileWatcher.DisableOperatorFileWatcher(); // don't update ops if file is written during save
-
-            RemoveAllSymbolFiles();
+            
+            // todo: this sounds like a dangerous step. we should overwrite these files by default and can check which files are not overwritten to delete others?
```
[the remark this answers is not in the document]
domportera: ya i figured this would remain a todo for roughly 86 years lmfao

Core/Resource/Model.cs
```diff
@@ -495,49 +532,52 @@ where type.IsSubclassOf(typeof(Instance))
 
         private readonly Regex _idFromNamespace = new Regex(@"(\{){0,1}[0-9a-fA-F]{8}_[0-9a-fA-F]{4}_[0-9a-fA-F]{4}_[0-9a-fA-F]{4}_[0-9a-fA-F]{12}(\}){0,1}",
                                                             RegexOptions.IgnoreCase);
-
-        private Symbol ReadSymbolFromFile(string filepath)
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static string ReplaceCharacter(string idString, char toReplace, char toReplaceWith)
```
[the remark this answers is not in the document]
domportera: oh no your gut is absolutely correct and i have no problem just yeeting it

Core/Resource/Model.cs
domportera: oh yeah your gut is absolutely correct and there's no noticeable difference to startup time and i have no problem just yeeting it

Core/Operator/Slots/Slot.cs
domportera: yes, this is the only place it is called:

```
public bool AddConnection(Symbol.Connection connection, int multiInputIndex)
        {
            var foundInstance = TryGetInstancesForConnection(connection, out _, out var sourceSlot, out _, out var targetSlot);

            if (!foundInstance)
                return false;
            
            targetSlot.AddConnection(sourceSlot, multiInputIndex);
            return true;
        }
```

Core/DataTypes/StructuredList.cs
domportera: yes it does, maybe it's new 🤷‍♂️

Core/Operator/Instance.cs
domportera: HARD agree on that one. I think it warrants being split into two methods - one for add and one for remove. I just translated what was there more or less

Core/Resource/Model.cs
```diff
@@ -442,51 +444,86 @@ private static void RegisterType(Type type, string typeName, Func<InputValue> de
...
+                                                 }
+                                             }
+                                             else
+                                             {
+                                                 Log.Warning($"Failed to load symbol {jsonInfo.Guid}");
+                                             }
+
+                                             return (success, symbol);
+                                         })
+                                 .ToList(); // execute and bring back to main thread
```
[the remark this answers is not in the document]
domportera: it's a dictionary because it needs to be queried in the `TryReadSymbolWithId` function. I can add some comments as well

Core/Resource/Model.cs
```diff
@@ -442,51 +444,86 @@ private static void RegisterType(Type type, string typeName, Func<InputValue> de
...
             {
-                for (int i = instanceTypes.Count - 1; i >= 0; i--)
-                {
-                    if (symbol.InstanceType == instanceTypes[i])
-                    {
-                        instanceTypes.RemoveAt(i);
-                        break;
-                    }
-                }
+                instanceTypes.Remove(symbol.InstanceType);
```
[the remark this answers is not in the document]
domportera: oh yeah i learned that one the hard way 😅 for this one though I could discern no good reason for it other than the fact that removing entries from a list by index is more efficient in reverse. By changing this to a hashset, that optimization is no longer required

Editor/Gui/Graph/Interaction/ExampleSymbolLinking.cs
```diff
@@ -10,15 +11,37 @@ public static class ExampleSymbolLinking
         public static void UpdateExampleLinks()
         {
             ExampleSymbols.Clear();
-            foreach (var symbolUi in SymbolUiRegistry.Entries.Values)
+
+            var symbolCollection = SymbolUiRegistry.Entries.Values;
+            var exampleText = "Example";
+            var examplesText = "Examples";
+
+            var potentialExamples = new List<Symbol>();
```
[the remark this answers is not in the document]
domportera: not sure I follow. I could definitely see making this a dictionary lookup by name to speed things up further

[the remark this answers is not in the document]
domportera: I re-optimized it and got the time down from 31.5k-33k ticks to 7.3k-9.6k ticks :)

Core/Resource/ResourceManager.cs
```diff
@@ -607,7 +607,7 @@ public uint CreateComputeShaderFromFile(string srcFile, string entryPoint, strin
             if (fileResource == null)
             {
                 fileResource = new ResourceFileHook(srcFile, new[] { newResource.Id });
-                ResourceFileWatcher._resourceFileHooks.Add(srcFile, fileResource);
+                ResourceFileWatcher._resourceFileHooks.TryAdd(srcFile, fileResource);
```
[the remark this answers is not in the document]
domportera: I'm pretty sure it's just the fact that parallel access to this dictionary led to duplicate keys being added, since between the time it checked if the dictionary had the key and the time it actually added it if it didn't, another thread may have already added that value

on second thought, I am reverting this change - I'd hate for this to "work" but not WORK if we ever end up threading that stuff

Core/Operator/Instance.cs
domportera: I did it

Editor/Gui/UiModel.cs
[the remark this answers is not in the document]
domportera: I solved this mainly by creating a local function

Editor/Gui/UiSymbolJson.cs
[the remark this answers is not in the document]
domportera: replaced instances of `FirstOrDefault` with `SingleOrDefault`

Editor/Gui/UiSymbolJson.cs
```diff
@@ -19,216 +19,192 @@
...
-            try
-            {
-                symbolId = Guid.Parse(mainObject["Id"].Value<string>() ?? string.Empty);
-            }
-            catch
-            {
-                Log.Error("Failed reading symbolUi id");
-                return null;
-            }
+            var vector2Converter = JsonToTypeValueConverters.Entries[Vector2Type];
```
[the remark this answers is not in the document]
domportera: addressed in pr #248

pixtur:
Merged! Thanks again for this awesome contribution

domportera:
happy to contribute :)

PR #248, opened 2023-04-08
Static type registration

domportera:
Referencing [this comment](https://github.com/still-scene/t3/pull/243#discussion_r1156430099). This PR specifically refers to [this commit](https://github.com/still-scene/t3/commit/af7d7657e1c82e8a850b2d99bb6bbe428866e99d), but ofc github is including all the previous changes too

I think this addresses your desire for those functions to be statically accessed in a safe way and makes the `Model` classes more comprehensible

edit: github doesn't like moving code to different files, but worry not. All this PR includes is moving "registration" functions to the static constructor of each `Model` class, and all of that logic to Model.TypeRegistration.cs partial class files. The `SymbolUiJson` class had its "converters"/writers (now "readers"/writers) pulled out into static variables initialized only once.


Editor/Gui/Graph/Interaction/ExampleSymbolLinking.cs
```diff
@@ -10,20 +12,62 @@ public static class ExampleSymbolLinking
...
+            foreach (var potentialExample in potentialExamples)
             {
-                var examples = SymbolUiRegistry.Entries.Values
-                                               .Where(c => c.Symbol.Name == symbolUi.Symbol.Name + "Example"
-                                                           || c.Symbol.Name == symbolUi.Symbol.Name + "Examples")
-                                               .Select(c => c.Symbol.Id)
-                                               .ToList();
-                if (examples.Count != 0)
-                    ExampleSymbols[symbolUi.Symbol.Id] = examples;
+                var gotSymbol = symbolsByName.TryGetValue(potentialExample.correspondingSymbolName, out var symbol);
```
pixtur, line 52:
```suggestion
                if(!symbolsByName.TryGetValue(potentialExample.correspondingSymbolName, out var symbol)
                     continue;
```

Would be just fine for me.

Editor/Gui/UiSymbolJson.cs
```diff
@@ -19,32 +19,34 @@
...
 {
-    public class SymbolUiJson 
+    public static class SymbolUiJson
     {
-        public JsonTextWriter Writer { get; set; }
-        public JsonTextReader Reader { get; init; }
+        private static readonly Func<JToken, object> JsonToVector2 = JsonToTypeValueConverters.Entries[typeof(Vector2)];
+        private static readonly Func<JToken, object> JsonToVector4 = JsonToTypeValueConverters.Entries[typeof(Vector4)];
+        private static readonly Action<JsonTextWriter, object> Vector2ToJson = TypeValueToJsonConverters.Entries[typeof(Vector2)];
+        private static readonly Action<JsonTextWriter, object> Vector4ToJson = TypeValueToJsonConverters.Entries[typeof(Vector4)];
```
pixtur, line 27: I'm kind of confused. Wouldn't those properties be initialized right at be beginning when `JsonToTypeValueConverters` might not event be initialized in `Model.cs`?  How does this work 🤔  ?

pixtur (approved):
This PR is pretty difficult to review on github because it contains to many other changes from the other PR. But I checked locally and it looks okay.

Editor/Gui/UiSymbolJson.cs
domportera: I'm pretty sure I'm relying on the runtime initialization order here. Since this class is never referenced before `UiModel`'s static constructor would be, these variables are guaranteed to have their intended values. My totally uninformed guess is that static values like this are initialized in the same way as a static constructor, so they wouldn't be evaluated until the `UiModel` references the `UiSymbolJson` class anyway

Editor/Gui/Graph/Interaction/ExampleSymbolLinking.cs
domportera: yeah I agree that's fine, I just think the former is more readable. six in one hand etc etc though

Core/Operator/Instance.cs
```diff
@@ -72,59 +72,70 @@ where typeof(IInputSlot).IsAssignableFrom(field.FieldType)
...
+              
+            // Get target Instance
             var targetInstance = compositionInstance.Children.SingleOrDefault(child => child.SymbolChildId == connection.TargetParentOrChildId);
-            ISlot targetSlot;
-            if (targetInstance != null)
-            {
-                targetSlot = targetInstance.Inputs.SingleOrDefault(e => e.Id == connection.TargetSlotId);
-            }
-            else
+            var gotTargetInstance = targetInstance is not null;
```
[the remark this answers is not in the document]
domportera: yeah it's pretty cool! I love how the language inches towards actual human-speak. some would call it syntactic sugar (which is often fair... though i like sugar :)) but in this case it's cool that it actually comes with a neat little unspoken guarantee

PR #257, opened 2023-05-15
Dependency separation for winforms

domportera:
I promise the size of this commit looks scarier than it is. The bulk of files touched just had simple WinForms calls replaced with interface calls.

Basically, I created `ICoreSystemUi` and `IEditorSystemUi` interfaces that are provided by a `SystemUi` singleton, as a sort of "Dependency Injection For Dummies" approach - that way we have the separation required for better modularity. Most all WinForms dependencies throughout Core and Editor projects are eliminated aside from keystrokes and possibly some cursor stuff.

I also managed to fix an annoying thing where when you drag the Editor window it turns tiny the first time (just by changing its default resolution to something more reasonable than 1x1 lol)

I also made the `SpaceMouse` class implement an interface (and reorganized/renamed the files appropriately) as its implementation was reliant on winforms too.

The only challenging work was in the re-organization of the window/device initialization, but it seems all good to me 👍 I think it's a little easier to wrap your head around (and harder to mess up) when looking at it from the main method. This stuff lives in the `ProgramWindows` class, and I've since extended the `AppWindow` class as that class seems to own the WinForms implementation.

Next steps here would be to 100% complete the separation of winforms (keystrokes, cursor, etc) and figure out an organizational scheme to ACTUALLY do proper dependency management (I don't see us needing a dependency injection framework or anything like that)


Editor/App/AppWindow.cs
```diff
@@ -109,15 +115,97 @@ public NoCloseRenderForm(string title) : base(title)
...
             rtv.Dispose();
             buffer.Dispose();
             swapChain.ResizeBuffers(3, form.ClientSize.Width, form.ClientSize.Height, Format.Unknown, 0);
-            buffer = Texture2D.FromSwapChain<Texture2D>(swapChain, 0);
+            buffer = Resource.FromSwapChain<Texture2D>(swapChain, 0);
             rtv = new RenderTargetView(device, buffer);
         }
 
         private bool _isResizingRightNow;
+
```
domportera, line 128: Just noticed I forgot to organize this stuff - but i guess now's my chance to properly ask you how you prefer variables and properties being organized in a class :)

PR #260, opened 2023-05-21
Separation into different C# projects

domportera:
this PR should clean itself up once my previous PR is pulled through (i assume)

This just refers to the "Cleanup/remove warnings" commit - no more big yellow/green lines - probably cleaned up a great deal of warnings with the GetWidth/Height methods around the project as well.

I figured removing the explicit redundant qualifiers (i.e. `SharpDx.Something.SomethingElse`) was ok now that this is an explicit implementation of sharpDX and imgui stuff only


SystemUi/SystemUi.csproj
```diff
@@ -0,0 +1,16 @@
...
+    <PropertyGroup>
+        <ImplicitUsings>enable</ImplicitUsings>
+        <Nullable>enable</Nullable>
+        <TargetFrameworks>net6.0;net6.0-windows</TargetFrameworks>
+        <RootNamespace>T3.SystemUi</RootNamespace>
+    </PropertyGroup>
+
+    <ItemGroup>
+      <Reference Include="System.Windows.Forms">
+        <HintPath>..\..\..\..\..\Program Files\dotnet\shared\Microsoft.WindowsDesktop.App\6.0.16\System.Windows.Forms.dll</HintPath>
```
pixtur, line 12: This looks kind of weird? What does this path point to?

domportera: oh that's weird. seems like it points to the local shared .net package - rider did this with nuget 🤷‍♂️

PR #263, opened 2023-05-21
remove win forms reference from SystemUi.csproj

domportera:
i heard you like pull requests so i figured one more would be more fun for everyone


pixtur:
More PRs, please!

PR #265, opened 2023-06-05
Better contrast for Blue operators

newemka:
I updated the color Blue for better reading while the mouse is hovering the operator


newemka:
I'm sorry I didn't expect the commit about Blob Ring and Sinform to be part of this pull request... Also I can't figure out how to cancel the pull request

pixtur:
No worries. I can close it here. I guess with the new theming system the contrast has improved.

PR #278, opened 2023-08-02
feat: add a node implementing Kanade flow across 2 images

ylmrx:


pixtur:
Merged on dev. Thanks for that awesome contribution! ♥

PR #280, opened 2023-08-07
fix: look-up resources from (includer') parent folder

ylmrx:
This would fix `#include` behavior whenever including headers relative to includers' folder.
This is the expected `#include` behavior, and would allow for more flexible ways to use these.
Also, this would allow user to embed some ready made shader libraries (such as lygia) easily.
Doesn't change current behavior beyond error messages.


pixtur:
Interesting. I have to look into this can make sure there are no side effects.

pixtur:
Thanks for this fix!

PR #281, opened 2023-08-08
add a GUI fullscreen setting for exported player

ylmrx:
- enable a new "Windowed Player" option in settings, allow to set the player behavior with a better UX. (CLI options are still working as before and prioritize over what's up in `projectSettings`)
- ... if this change is good, more settings could probably be enabled (res ? vsync ?)
- also fixes a few cosmetic around player.exe (window title, some text)


ylmrx:
image

pixtur:
That's a really good idea and it's totally worth an option.

VisualBoy:
It's possible to add  an option to allow Windowd Player resizable?

`AllowUserResizing = false`

PR #284, opened 2023-08-23
fix: new fullscreen settings for many-screens setups

ylmrx:
- adds some index options to specify which monitor should show fullscreen windows
- remove the confusing "swap displays" toggle

- the 2 new options can be `Enums` too, thus a bit tricky to populate comfortably (advise if needed)
- should we re-address (basically toggle it off and on) the screen when setting is updated in order to update immediately ? or should this be a user action ?

<3

### DIFF
--- a/Editor/App/ProgramWindows.cs
+++ b/Editor/App/ProgramWindows.cs
@@ -11,7 +11,9 @@ using T3.Core.Logging;
 using T3.Core.Resource;
 using T3.Editor.Gui;
 using T3.Editor.Gui.UiHelpers;
+using T3.Editor.Gui.Windows.Layouts;
 using T3.Editor.SystemUi;
+using T3.Editor.UiModel;
 using T3.SystemUi;
 using Device = SharpDX.Direct3D11.Device;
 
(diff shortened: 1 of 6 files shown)


ylmrx:
it (should) look like this 🌞 :

image

(vlc for ref) : 

image

pixtur (commented):
looks good to me, However the WindowsForms abstraction would be very nice.
I could look into that, if you want me to.

pixtur:
Hi @ylmrx could you rebase the PR to the latest dev commit?

ylmrx:
there you go @pixtur !

I set it to hide the mainmenubar and the timeline when in focus mode too.

ylmrx:
My focus mode fix was breaking, but i figured the confusing imgui.menuitem strange behavior...
We can now have an "checkmarked" menu entry , which still take its bool return into account by removing "ref", now the if gets correctly executed, and the focus mode works as intended from both the key shortcut, and the menu entry (which also get its correct status)

pixtur:
Thanks for your contribution!

PR #286, opened 2023-09-02
Symbol Change (right click->Change Symbol)

teadrinker:
Implementation info, steps (for each selected SymbolChild):
(1) Create a new SymbolChild
(2) Loop through in/out and match connections/data by name and type and copy/connect to newly created
(3) If conversion was lossless, remove original SymbolChild
(4) Select new SymbolChild

For UI, I modified the SymbolBrowser to be able to return the target Symbol in a callback instead of creating it directly

Reason for this PR:
I felt it was slow and error prone to manually reproduce properties and connections when switching between different versions or implementations of operators (for instance when using "Duplicate as new type..." to edit a built-in operator)

(I have almost no experience with the codebase so I probably missed some stuff, or messed up some stuff!)

### DIFF
--- a/Editor/Gui/Graph/GraphCanvas.cs
+++ b/Editor/Gui/Graph/GraphCanvas.cs
@@ -828,6 +828,30 @@ namespace T3.Editor.Gui.Graph
             }
 
             ImGui.Separator();
+
+            if (ImGui.MenuItem("Change Symbol", someOpsSelected))
+            {
+                var startingSearchString = selectedChildUis[0].SymbolChild.Symbol.Name;
+                var position = selectedChildUis.Count == 1 ? selectedChildUis[0].PosOnCanvas : InverseTransformPositionFloat(ImGui.GetMousePos());
+                SymbolBrowser.OpenAt(position, null, null, false, startingSearchString, symbol =>
+                    {
+                        var nextSelection = new List<SymbolChild>();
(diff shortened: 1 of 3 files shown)


pixtur:
Wow, that looks like an incredible PR! The only thing that pops into my mind is the handling of the Undo-Stack. Did you try to undo the replacement? From looking at the code, it would probably take multiple undo-commands which will lead to an inconsistent state.

There are two things that could be changed ideally:

1. You could simply collect the commands and then create and add a MacroCommand to the UndoStack.
2. You could insert the replace operator into the ConnectionMaker class. This implements "operations" of which the replacement could be one.

Let me know what you think. I could also take over from here.

pixtur:
Ups. Merged by accident. Sorry for the confusion.
I noticed that you did your changes on the master branch. Would it be possible to create a PR (or change this one) to dev?

teadrinker:
Ah, yeah you're right, the multi-step undo would be confusing for the user!

I have updated the branch ([teadrinker:feat/change-symbol](https://github.com/teadrinker/tooll3/tree/feat/change-symbol)) so that it uses MacroCommand. I also merged in dev.

teadrinker:
(should I create a new PR since this is closed btw?)

pixtur:
I noticed you already did! Thanks a lot!

PR #287, opened 2023-09-09
Symbol Change (right click->Change Symbol) (now using base:dev)

teadrinker:
It seems I had to create a new branch and cherry pick the edits for git not to be confused...
Uses MacroCommand and seems to work as expected (but probably needs more testing naturally)


pixtur:
Awesome! Thanks a lot ❤️

PR #292, opened 2023-09-11
Draw Mesh Cel Shading

newemka:
A gradient driven shader that allow Cel shading but not only. 
There is an example to show how to use the shader and add outline. 
The shader is a hack of DrawMeshHatched


pixtur:
Awesome! Thanks!

Issue #288, opened 2023-09-09
(minor/usability) clicking "I know what I am doing" does not result in expected action


teadrinker:
When opening a lib operator, it shows a warning (which is great).
Then the next time you to open a lib operator, you get straight inside.
 
However I feel the expected behaviour is that if I click "I know what I am doing", I should not have to repeat the action manually.

pixtur:
I know the UX is not ideal, but I quickly looked into this and couldn't find an easy fix for this.

ylmrx:
I think the repeated "I know what I'm doing" is the desirable behavior (but we could have a user setting "I really know what I'm doing", which disable all of these warnings)

wdyt ?

pixtur:
Fixed with PR

PR #293, opened 2023-09-16
fix: rework of the 'i know what i'm doing' pop-up

ylmrx:
This solve #288 

- we get inside child upon clicking the button (you had to double click the op again before)
- "don't ask again" now correctly reflect and update the Warn setting...
- ... and can be canceled too (the settings is discarded then)
- cosmetics


pixtur:
Thanks. It's a small thing, but it really helps!

PR #294, opened 2023-09-16
Fixing CelShader path

newemka:
I accidentally created a ReSSources (with two S) folder where the HLSL file was. Now the HLSL file is in Resources folder and I updated its paths in the operator


pixtur:
Thanks!

PR #296, opened 2023-09-19
Fix T3 crash if Copy an empty Console

newemka:
If you hit the Console's Clear button then hit Copy button T3 will crash. Fixed by adding a "Console cleared" log entry.


pixtur:
Ha! That's a good work around. But it would be better to fix the actual crash reason.
I can look into it.

PR #306, opened 2023-10-19
Lock Output Canvas in place

newemka:
A user setting to avoid moving and scaling the Output canvas


newemka:
A better solution has been already pushed

Issue #301, opened 2023-10-10
(Minor issue) Minimap don't show inputs/outputs


teadrinker:
v.3.7.1 Dev

a bit confusing when graph looks different in the minimap

PR #309, opened 2023-10-21
fix: enable in and outs in the minimap. (fix issue #301)

ylmrx:


pixtur:
Hi @ylmrx, your code looks fine. But sadly that would basically mean, that the graph is rendered twice. I believe the performance impact would be too large. I personally think that it's perfectly valid that the minimap is an abstraction of the actual graph.

PR #314, opened 2023-11-13
TransformImage2 (Proposal) 

newemka:
Here is a proposal for a new version of the "TransformImage" tool. I removed the mirror setting and added a TextureAddressMode" input parameter connected to "wrap" input of the _ImageFxShaderSetupStatic. 
I duplicated the HLSL file and accordingly modified it.
I created an example symbol in order to describe the difference between both operators.
Descriptions have been updated as well.


pixtur:
Nice. Would it be possible to merge it with the existing [TransformImage] ? It would be awesome if we could keep the number of lib ops as small as possible (or necessary).

newemka:
Sure I will do it ASAP

newemka:
In fact I don't like that TextureAddressMode can't be animated like a normal enum. I'll update TransformImage's HLSL code in order to reproduce the different wrapping modes

newemka:
Ready to merge

pixtur:
Thanks a lot ♥

PR #316, opened 2023-11-25
Redo sharpdx math replacement

domportera:
Tested against the issues presented, and all seems to be just dandy. im sure you'll want to test on your end before pulling this in though.

I wouldn't call this a complete wipe of sharpDX math, but it's certainly a lot. next id like to make int vectors uniform and have some stuff in progress for that


pixtur:
Thanks a lot for that great work!

PR #323, opened 2023-12-07
Aspect Ratio fixed

newemka:
I also added a Snowflake preset and edited the description a bit. There was also a weird behavior with steps, 0 was giving a black screen. I made a small change in the HLSL and now the setting is clamped from 1 to 666.


pixtur:
Thanks a lot!

newemka:
> Thanks a lot!

My pleasure!

PR #325, opened 2023-12-12
fix changed color-out-of-bounds behavior

domportera:


pixtur:
I was kind of confused why we're not using `Mathf.Max()` but I assume it doesn't make much of a difference in this case.

domportera:
> I was kind of confused why we're not using `Mathf.Max()` but I assume it doesn't make much of a difference in this case.

yeah im not sure, Math.Max has a lot of overloads for types now, and im fairly sure one is float. if it's actually double may be worth switching

PR #327, opened 2023-12-14
Shader resources

domportera:
I've better compartmentalized Shader compilation and resources generated from them so it should be a great deal easier to pull out DirectX from this process in the future. Much work to be done on that front, but this is a good jumping-off point. Lots of redundant code has been removed.

I haven't yet reorganized the actual file locations for scripts associated with this functionality. I also think further work could be done to generalize the shader operators, but that's a task for another day.

There's also some nonsense in here you might see from when I was poking around at what I should do next for the Great T3 Refactor 😅 but just a little I promise

Added bonus: shared shader bytecode!


pixtur:
Awesome work!

domportera:
> Awesome work!

thanks!!

PR #328, opened 2023-12-17
NDI : Improve Upload performances

mrvux:
Add a function to properly copy image memory (also handles unusual paddings)

Replace old copy function to the new one.


pixtur:
I love you!

PR #332, opened 2023-12-21
fix: do not let the teach mode override the pre-connected inputs

ylmrx:


pixtur:
Excellent improvement. Thanks!

PR #339, opened 2024-01-06
fix: lines after the appearance of a reset button would get mis-aligned

ylmrx:


ylmrx:
This is a very hackish solution to a very cosmetic issue, and is still sub-optimal.
Maybe more tables could prevent such tiny render bugs.

pixtur:
Thanks!

PR #340, opened 2024-01-06
fix: prevent uiscale from being zero

ylmrx:


ylmrx:
When entering a uiscale, if the user starts with '0' or '.', Tooll will parse it as a 0, and the UI will break immediately (the field lose the focus, it can't be edited anymore, the 0 value is saved to userSettings and has to be fixed manually there)

This patch prevent such situation from happening by forcing it back to some minimal value when it get to close to zero and let the edit happen entirely.

Editor/UiContentUpdate.cs
```diff
@@ -25,9 +25,11 @@ public static void RenderCallback()
         CursorPosOnScreen = new Vector2(cursorPos.X, cursorPos.Y);
         IsCursorInsideAppWindow = ProgramWindows.Main.IsCursorOverWindow;
 
-        // Update font atlas texture if UI-Scale changed
         if (Math.Abs(UserSettings.Config.UiScaleFactor - _lastUiScale) > 0.005f)
         {
+            // Prevent scale factor from being "actually" 0.0
+            if (UserSettings.Config.UiScaleFactor < .1f) UserSettings.Config.UiScaleFactor = .1f;
```
pixtur, line 31:
```suggestion
            if (UserSettings.Config.UiScaleFactor < 0.1f) UserSettings.Config.UiScaleFactor = 0.1f;
```

pixtur (changes requested):
Thanks for you fix. That's a excellent and straight forward contribution.
With the T3 code base we try to follow the .net coding style guideline, It would be great if you could insert the leading zeros for consistency's sake.

pixtur:
Thanks!

PR #344, opened 2024-01-18
Add description to [Time] and [Multiply] operators

alixcarolan:


pixtur:
I'm sorry. Sadly, I was working on the same descriptions as an example which caused a merge conflict.
Could you update this PR ?

PR #345, opened 2024-01-18
Update ShardNoise and Worley

newemka:
Bias and Gain for ShardNoise
Now we can use the colors' alpha channel.
Added Offset parameter.

I reordered WorleyNoise's parameters


pixtur:
Thanks!

PR #357, opened 2024-01-27
New Modes for StarGlowStreaks

newemka:
I also experimented a Quality parameter based on MipMaps. Bonus a T3 logo OBJ with clean UVs.
StarGlowStreaksModes


newemka:
DrawBillboards update. Points too close to the camera are fading out if the Bool parameter is true
https://github.com/tooll3/t3/assets/3755089/5ddd8b00-6944-4170-92c8-aca43274f393

newemka:
ToneMapping

1. image before ToneMapping
2. ToneMapping default parameters (Now the glow looks much better)
3. GammaCorrection enabled with its default value (according to me it makes the dark areas too bright)
4. Gamma value tweaked to please my eyes (Maybe it's not a correct way to perform a gamma correction, but I like the result)

PR #360, opened 2024-01-29
Replaced int3 by int4 for face/triangle indices

newemka:


pixtur:
Thanks for the hard work!

newemka:
Team work! Thanks for the support!

PR #362, opened 2024-02-01
Better feedback from its UI and Gizmo

newemka:
### MeshProhectUV :
 Now the gizmo and UI are working together and make the operator more user friendly


newemka:
image

PR #363, opened 2024-02-01
string operator IndexOf

leon196:
added a simple "index of" operation for string.
useful for checking if a name has a particular character.

### DIFF
--- a/Operators/Types/lib/string/IndexOf.cs
+++ b/Operators/Types/lib/string/IndexOf.cs
@@ -31,14 +31,14 @@ namespace T3.Operators.Types.Id_fc0a5e68_9915_4323_b2a4_2491fa5d59a9
                 return;
             }
 
-            Index.Value = originalString.IndexOf(searchPattern);
-            if (string.IsNullOrEmpty(searchPattern))
+            try
             {
-                Index.Value = -1;
-                return;
+                Index.Value = originalString.IndexOf(searchPattern);
             }
-            string originalString = OriginalString.GetValue(context);
-            Index.Value = originalString.IndexOf(searchPattern);
+            catch (Exception)
+            {
+                Log.Error($"'{originalString}' or '{searchPattern}' is incorrect", this);
+            }
         }
 
         [Output(Guid = "4bb4bb23-4c3f-4d7d-9dab-c37ac63dd1c9")]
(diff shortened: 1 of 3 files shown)


Operators/Types/lib/string/IndexOf.cs
```diff
@@ -0,0 +1,39 @@
...
+        public readonly InputSlot<string> SearchPattern = new();
+
+        public IndexOf()
+        {
+            Index.UpdateAction = Update;
+        }
+
+        private void Update(EvaluationContext context)
+        {
+            string searchPattern = SearchPattern.GetValue(context);
```
pixtur, line 26:
```suggestion
            string searchPattern = SearchPattern.GetValue(context);
            string originalString = OriginalString.GetValue(context);
            if (string.IsNullOrEmpty(searchPattern))
            {
                Index.Value = -1;
                return;
            }

            Index.Value = originalString.IndexOf(searchPattern);
```

Although the code looks valid and clean there is a small but important detail hidden here:

If you exit early, before getting the value for `OriginalString.GetValue(context);` this input will always stay dirty. This means that every Operator right of it will stay dirty and can't be cached.

For that reason, gathering all inputs before existing is the better thing to do here. 

Another detail: With string operators I'm personally never sure when it throws exceptions. In your case, you should also check `originalString` for null before call `IndexOf`. Otherwise this operator will throw an exception if connected to a null string. (Currently by design throwing and exception will exist the exception - which looks like a crash to most users).

If in doubt add a try/catch statement around the functions in question, just to be on the save side.

pixtur (changes requested):
Thank a lot for this contribution! An [IndexOf] operator would be great, but I think we should make it bullet proof first.

Operators/Types/lib/string/IndexOf.cs
leon196: I see, interesting! Thanks for the explanation. I've added the suggestions.

Operators/Types/lib/string/IndexOf.cs
```diff
@@ -24,13 +24,21 @@ public IndexOf()
...
                 return;
             }
-            string originalString = OriginalString.GetValue(context);
-            Index.Value = originalString.IndexOf(searchPattern);
+
+            try
+            {
+                Index.Value = originalString.IndexOf(searchPattern);
+            }
+            catch (Exception)
```
pixtur, line 38: Sorry for the confusion. In that case, just checking for null in the line above would have been completely sufficient :-)

PR #364, opened 2024-02-02
fix typo in transform vertices shader

leon196:
It seems that there is a typo in the mesh-TransformVertices.hlsl file.
line 33: it interpolates between normal and transformed tangent.

PS:
I'm trying to only commit this change, but it seems that it includes all the commits of my fork?
Sorry if I messed up the commits in this pull request.


pixtur:
Okay. It only listed on changed file which looked fine to me.

PR #367, opened 2024-02-07
An equirectangular rendering node

leon196:
A node that takes a draw command and output an image color and depth with equirectangular mapping.

image

Each face of the cubemap is unwrapped and remapped as an equirectangle

image

Based on Bartosz answer from this stackoverflow link:
https://stackoverflow.com/questions/34250742/converting-a-cubemap-into-equirectangular-panorama


pixtur:
Wow. Thanks a lot of this contribution!

PR #372, opened 2024-02-11
Adding Color factors to SamplePointAttributes

newemka:
It still needs some adjustments
Dealing with the Alpha channel is always confusing. 
image


pixtur:
Thanks!

PR #375, opened 2024-02-16
Add a folder to userSettings for image sequence render

newemka:
Same behavior as the Video render window.
It will create a default folder to render image sequence if the user didn't select a folder.
If the user select an other folder to render image sequences, T3 will remember it.


pixtur:
Thanks!

PR #376, opened 2024-02-17
Multi-project operators, slot optimizations, dependency management, etc

domportera:
the MEGA PR

some work remains to be done, but most things should still work quite nicely as far as i can tell


pixtur:
This has been merged into the main branch with CLI tools.

PR #377, opened 2024-02-17
Fix ColorHSB parameter's  behavior

newemka:
RandomizePoints.hlsl was using RgbToLCh function, I replaced it by rgb2hsb and hsb2rgb so ColorHSB parameter is behaving as expected. 

https://github.com/tooll3/t3/assets/3755089/0abf9f4b-0121-4961-84ce-ab26ff267419


pixtur:
Good catch!

PR #378, opened 2024-02-18
Equirectangle update

leon196:
Updated resolution parameter with an enumerator.
Tried to fix seam issues and maximum resolution possible.
Did change the way it works, using render targets instead of slicing viewport, allowing better face resolution of the cubemap.
Unfortunately the depth buffer had seam with this technique and is very costly in performance for SSAO.
For the moment depth buffer is disabled.

image


leon196:
something is wrong at the moment, I have to fix the things before doing an actual pull request.

PR #386, opened 2024-02-29
TransformImage translation on X according to mouse movement

newemka:
It's a small thing but it feels so much better to see the image moving in the same direction as the mouse


pixtur:
Thanks for the cleanup!

PR #388, opened 2024-03-03
DrawBillboards RandomRotate Default value set to 0 instead of 180

newemka:
When using Orientation parameter: PointRotation
We expect to get the points orientation defined before DrawBillBoards operator
With RandomRotate default value set to 180, this parameter is hidden by default like in the screenshot 
image
And the result is an unexpected rotation of the points.


pixtur:
I already fixed the shader so changing the default settings here is no longer necessary.

PR #389, opened 2024-03-03
Fix [BuildAsciiFontSorting] file resolution

domportera:
This will likely be a recurring theme as we polish this release and find places that files aren't found - but this is how the problem is fixed!


pixtur:
Wow. Thanks a lot!

PR #390, opened 2024-03-06
AAC audio codec for video Export

newemka:


pixtur:
Super cool, that you looked into that. Much better than mp3.

PR #394, opened 2024-03-13
Deform Mesh

newemka:
There are two versions of DeformMesh 
DeformMesh2 only do twist


pixtur:
Awesome contribution!

PR #400, opened 2024-03-22
feat: a PDF reader operator and its example

ylmrx:
the example pdf file was made by @newemka


ylmrx:
The imagemagick library at work here still require to have Ghostscript installed in order to convert the pdf files.

pixtur:
I will reopen this once we switched to the new package management on main

PR #401, opened 2024-03-23
feat: adding a LaTeX formula render operator

ylmrx:


pixtur:
I will reopen this once we switched to the new package management on main

PR #403, opened 2024-03-25
Feature/spline points update

newemka:
[SplinePoints] had not color and no stretch values, I added default values for these point attributes.
The points orientation was flipped, [ExtrudeCurves] was giving inside out meshes, I've tried to fix this.

I added a [SplinePointsExample]. 

An issue remains, for example  [RandomizePoints] has zero effect on the points of [SplinePoints], I don't understand why. 

I added a [CameraExample] and added view presets for the camera.


pixtur:
Thanks a lot!

PR #404, opened 2024-03-26
feat: add an OscOutput operator and its example

ylmrx:
This is a new operator, counterpart to OscInput
It _only_ supports strings and float numbers, it tries to fail gracefully on bad location and failure to connect.
Error and state check is grunty, please advise if I'm doing it too wrong.

### DIFF
--- a/Operators/Types/lib/io/osc/OscOutput.cs
+++ b/Operators/Types/lib/io/osc/OscOutput.cs
@@ -1,15 +1,15 @@
 using T3.Core.Operator;
 using T3.Core.Operator.Attributes;
 using T3.Core.Operator.Slots;
-using T3.Core.Logging;
 using System.Net;
 using Rug.Osc;
 using System;
 using T3.Core.Utils;
+using T3.Core.Operator.Interfaces;
 
 namespace T3.Operators.Types.Id_4e99da86_482f_4037_8664_b2371526d632
 {
-    public class OscOutput : Instance<OscOutput>
+    public class OscOutput : Instance<OscOutput>, IStatusProvider
     {
         [Input(Guid = "98f38caa-4f79-425c-ab46-22d7dbe62978")]
         public readonly InputSlot<string> IpAddress = new InputSlot<string>();
@@ -27,9 +27,11 @@ namespace T3.Operators.Types.Id_4e99da86_482f_4037_8664_b2371526d632
         public readonly InputSlot<float> Number = new InputSlot<float>();
 
         [Input(Guid = "4931217b-e30d-42a6-9cda-bc49a8fd8d67", MappedType = typeof(OscTypes))]
-        public readonly InputSlot<int> OscValueType = new ();
+        public readonly InputSlot<int> OscValueType = new();
 
         [Output(Guid = "a6679d6c-fc34-4588-ab20-5079ad8f8a03")]
+        public readonly Slot<T3.Core.DataTypes.Command> Result = new Slot<T3.Core.DataTypes.Command>();
+
         public OscOutput()
         {
             Result.UpdateAction = Update;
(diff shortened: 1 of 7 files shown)


Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+using System.Net;
+using Rug.Osc;
+using System;
+using T3.Core.Utils;
+
+namespace T3.Operators.Types.Id_4e99da86_482f_4037_8664_b2371526d632
+{
+    public class OscOutput : Instance<OscOutput>
+    {
+        private bool _connected;
```
pixtur, line 14: Please always move private members to the end for the class.

Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+        private IPAddress _address;
+        private int _port;
+        private float _num_value;
```
pixtur, line 17:
```suggestion
        private float _numericValue;
```
Please follow the .net c# naming conventions and avoid short forms.

Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+        private string _str_value;
+        private OscSender _sender;
+        public OscOutput()
+        {
+            Result.UpdateAction = Update;
+        }
+
+        private enum OscType
```
pixtur, line 25:
```suggestion
        private enum OscTypes
```

Please move to bottom and use plural for enumerations.

Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            Number,
+            String
+        }
+        private void Update(EvaluationContext context)
```
pixtur, line 30: Please leave blank line after closing `}`

Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+                }
+            }
+            catch (Exception e) {
+                Log.Error("Connection failure: " + e);
+            }
+            _connected = false;
+        }
+
+        [Input(Guid = "98f38caa-4f79-425c-ab46-22d7dbe62978")]
+        public readonly InputSlot<string> Address = new InputSlot<string>();
```
pixtur, line 121: Would `IpAddress` be a better name?

Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+            String
+        }
+        private void Update(EvaluationContext context)
+        {
+            var location = Location.GetValue(context);
+            var num = Number.GetValue(context);
+            var str = String.GetValue(context);
+            var oscType = OscValueType.GetEnumValue<OscType>(context);
+
+            CheckAddr(context);
```
pixtur, line 37:
```suggestion
            if(!TryGetValidAdress(IdAddress.GetValue(context), out var problem, out var address) {
               _lastErrorMessage = problem;
            }
```


I would have expected a `CheckXYZ` method to return a boolean. 
I much prefer the `TryGetXYZ` pattern being as static as possible.

Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+                    if (num != _num_value)
+                    {
+                        // Log.Info("Sending:" + num);
+                        OscMessage message = new OscMessage(location, num);
+                        if (message.Error == OscPacketError.None)
+                        {
+                            _sender.Send(message);
+                            _num_value = num;
+                        }
+                        else Log.Error("failed to build the OSC message (float): " + message.ErrorMessage);
```
pixtur, line 55: Please start messages with an uppercase letter.

Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+
+namespace T3.Operators.Types.Id_4e99da86_482f_4037_8664_b2371526d632
+{
+    public class OscOutput : Instance<OscOutput>
+    {
+        private bool _connected;
+        private IPAddress _address;
+        private int _port;
+        private float _num_value;
+        private string _str_value;
```
pixtur, line 18:
```suggestion
        private string _stringValue;
```

Operators/Types/lib/io/osc/OscOutput.cs
```diff
@@ -0,0 +1,142 @@
...
+        [Input(Guid = "d5e9e9be-093b-4d57-9070-9b2cf33aa45b")]
+        public readonly InputSlot<string> String = new InputSlot<string>();
+
+        [Input(Guid = "827c7ae2-c129-4d04-a715-328c0a86bf8a")]
+        public readonly InputSlot<float> Number = new InputSlot<float>();
+
+        [Input(Guid = "4931217b-e30d-42a6-9cda-bc49a8fd8d67", MappedType = typeof(OscType))]
+        public readonly InputSlot<int> OscValueType = new ();
+
+        [Output(Guid = "a6679d6c-fc34-4588-ab20-5079ad8f8a03")]
```
pixtur, line 139: Outputs should be at the beginning of the file. (I know... it's strange. It's not my convention. But at least it is one).

pixtur (changes requested):
That's an excellent PR! Most of my comments are nitpicky about code-style and conventions. But there are three things that should be done:

1. The Op should implement the Dispose-Pattern and probably disconnect when disposing. (If not, it might happen that the op keeps an connection open).
2. Another problem might be that too many open connections will lead to performance issues. I'm not sure if this is true but OscInput uses a shared OscConnectionManager to pool and reusing connections. Although this optimization might be quite involved. I think we can live without that here.
3. Please implement `IStatusProvider`.  This will show a warning icon whenever the op has a problem (like a missing connection or an invalid IP address). Check [AbletonLiveSync] for an example.

ylmrx:
(disregard latest commit, involving very experimental/wip/poc regarding text support for OscInput, i wanna keep this change focused on OscOutput), switching it to draft, as i might be noisy in between)

ylmrx:
> That's an excellent PR! Most of my comments are nitpicky about code-style and conventions. But there are three things that should be done:

Thanks ! :)

>     1. The Op should implement the Dispose-Pattern and probably disconnect when disposing. (If not, it might happen that the op keeps an connection open).

done.

>     2. Another problem might be that too many open connections will lead to performance issues. I'm not sure if this is true but OscInput uses a shared OscConnectionManager to pool and reusing connections. Although this optimization might be quite involved. I think we can live without that here.

The conManager is built with a server purpose in mind, and won't be much use for us here.

I'm not sure a threaded model would help us in this context, as there's no error check on the `OscRug.foobar.Send()` (we're UDP), and a given operator should address its whole traffic (all distinct IPs/port/location)

We can check a later stage, for a threaded model, if we get performance issue, but i think we're fine here.

On another, perf-related note, the operator gives a tiny glitch once in while, but seems to be related to the refresh rate of Tooll (and phase with touchosc frame rate (where you can't disable vsync) : it completely disappear when I disable vsync (in tooll) and it stays well under 1ms frame-time).

This is a general issue with OSC client (TouchOSC does that too (this is why many standalone sequencer/drumbox were attempted and failed (15ms is barely noticeable for eye, but sound like a clumsy drummer for a sequencer)

>     3. Please implement `IStatusProvider`.  This will show a warning icon whenever the op has a problem (like a missing connection or an invalid IP address). Check [AbletonLiveSync] for an example.

Done!

pixtur:
Awesome!

PR #405, opened 2024-03-26
MeshFacesPoints

newemka:
Set points at the "Inscribed Circle Center" of the mesh's faces, the W attribute gets the radius


pixtur:
Thanks!

PR #410, opened 2024-03-30
Feature/spline points sample color

newemka:
[SplinePoints] will use the color attribute from [Apoint]

Also fixes a bug that was reported about [Apoint] no being rendered in a exported executable. Due to the buffer not being updated when the application starts.


pixtur:
Thanks!

PR #412, opened 2024-04-01
TextOutlines2

newemka:
This is a outline variation of [Text] 

https://github.com/tooll3/t3/assets/3755089/b92161fb-c511-47c4-9fc6-f845098a7b97


pixtur:
Oh wow. That is a an awesome contribution! Thanks a lot!

PR #413, opened 2024-04-03
Feature/brings back midi to export

newemka:
I'm not sure it's the best way to bring back midi support to executable, but it works 
I had to revert Text Points commit (ee05a26fdc7996205d5cc2a943ff3d2847cae468) in order to test the standalone


pixtur:
Thanks!

PR #415, opened 2024-04-05
Feature/ An Operator to visualize the UV map of a mesh 

newemka:
It's based on [BlendMeshVertices], but instead of blending from one mesh to an other it blends from the mesh to its UV coordinates.

While testing it on [CylinderMesh] it helped to see that all its UV islands were not inside the UV map area then I managed to edit CylinderMesh.CS in order to fix this. 

This video describes a use case and shows [CylinderMesh] UV map:
https://github.com/tooll3/t3/assets/3755089/31855c9c-d570-4bc7-aa9d-45a8fcf5a825

Here is the corrected UV map for [CylinderMesh]
CynlinderMeshUVs


pixtur:
Great contribution! Very helpful as a debugging tool.

PR #417, opened 2024-04-08
[BoundingBoxPoints]

newemka:
It iterates through the source points, excluding those with invalid W values (-nan(ind)), and calculates the minimum and maximum bounds in each dimension (x, y, z). These bounds define the eight points of the bounding box: the minimum and maximum corners of the box. Additionally, it computes the middle point of the bounding box by averaging the minimum and maximum bounds

image


pixtur:
This is a super cool operator. But I think we should refactor the implementation. I started a discussion thread here: https://discord.com/channels/823853172619083816/1238088698374131733

PR #418, opened 2024-04-08
fix: spawn the MidiConnectionManager along with MidiOutput

ylmrx:


pixtur:
Thanks for the contribution

PR #419, opened 2024-04-15
fix: dupe CtrlValue in debug message, WasHit issued with connected Control too

ylmrx:
- Duplicated controller value with PrintLogMessage enabled
- WasHit wasn't issued with a wired Control parameter
- useless Control.Value update (happens only in the `if(_teachingActive)` section)


pixtur:
Thanks. Combined, these incremental improvements really make a difference!

PR #420, opened 2024-04-16
Feature/spring operator

newemka:
The [Spring] operator is a port of this [spring movement function](https://twitter.com/i/status/1148297551931572224)

https://github.com/tooll3/t3/assets/3755089/4b21f6f3-0994-4d54-bc37-ec14689ae970


pixtur:
I love it! We should add a small comment to [Damp] so people will find this alternative interpolation.
In the long term, it might also be cool to have something like [DampPoints]. However this will involve a secondary structured buffer to store the velocities. So it's probably going to be more involved.

newemka:
I'm glad you like it! I'll will make its vec2 and vec3 siblings. 
For this I think I'll need to have the spring function somewhere else in order to use it multiple times. 
Should I add to MathUtils.cs ?

PR #423, opened 2024-04-17
feat: adds lib.math.float.TryParse and lib.math.int.TryParseInt

ylmrx:
pretty self-explanatory, tries to parse some text to a float number.
If it won't parse it fails over to the default value you can set.


pixtur:
Thanks for the contribution. What do you think about renaming it to StringToFloat to make more consistent with other ops?

PR #424, opened 2024-04-17
fix: video playback duration value glitching, fix VideoPlayExample

ylmrx:


ylmrx:
⚠️ I might be doing it as I'm unsure if `_hasUpdatedTexture = false;` removal has no side-effects (I didn't get any)
The Duration was glitching alternating -1 and correct Duration, when the Update was called twice during the same frame (this happens when video file is using <60fps)

ylmrx:
I also adjusted few stuffs in PlayVideoExample, it now works correctly at least with files found in `Resources\common\test-footage`

pixtur:
Good catch. I frequently check the context.LocalTime to test if the operators has already been updated in the current frame.  The _hasTextureUpdated was an attempt to only execute some compute shaders if the texture actually changes. This is relevant when apply stuff like motion vector extraction to videos with framerates like 25fps or if the current display rate is not 60 fps (like note books with 144 FPS refresh rate).

PR #426, opened 2024-04-27
feat: [ConvertPointAttributes] operator

alixcarolan:
A relatively low-level shader operator for simple conversions from one point attribute into another. Not super flashy on its own, but has potential when stacked a few times in a row or combined with other point operators!

The operator is also documented in-app.


alixcarolan:
I think this op would be more useful with a BiasAndGain control. I'm too busy to work on one at the minute but I should be able to get to it soon enough...

pixtur:
What is the current status of this PR?

alixcarolan:
I was going to work on making this op more intuitive and flexible before reopening the PR... but it seems like I forgot about it instead 🙃

I think I overwrote my dev environment too so I can't work on it anymore even if I wanted to.

As is, it still provides functionality that isn't currently present in T3 outside of CustomPointShaders, so I'll just reopen it.

PR #428, opened 2024-05-10
Adds PlayAtlas and SnapToMultiple operators

anosci:
An operator for turning Atlases into animations, for use with GIFs, sprite sheets, etc. For a better user experience, I've also included SnapToMultiple, which will return the lowest divisor of a number that has a modulus of 0. 
playatlas

### DIFF
--- a/Operators/Types/user/community/SnapToMultiple.cs
+++ b/Operators/Types/user/community/SnapToMultiple.cs
@@ -1,0 +1,38 @@
+using T3.Core.Operator;
+using T3.Core.Operator.Attributes;
+using T3.Core.Operator.Slots;
+
+namespace T3.Operators.Types.Id_c3090c65_194a_4f95_9b70_d003f54103f7 
+{
+    public class SnapToMultiple : Instance<SnapToMultiple>
+    {
+        [Output(Guid = "87a311ce-238c-472c-b43e-e4ed5268bbc5")]
+        public readonly Slot<int> Result = new();
+
+        public SnapToMultiple()
+        {
+            Result.UpdateAction = Update;
+        }
+
+        private void Update(EvaluationContext context)
+        {
+            var v = Value.GetValue(context);
+            var mod = Mod.GetValue(context);
+            if (mod == 0){
+                return;
+			}
+            for(int goodvalue=mod;goodvalue>0;goodvalue--){
+				if(v%goodvalue == 0){
+					Result.Value = goodvalue;
+					return;
+				}
+			}
+        }
+        
+        [Input(Guid = "3bf35135-8fb1-46d3-95ea-008eded67060")]
+        public readonly InputSlot<int> Value = new();
+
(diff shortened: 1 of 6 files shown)


Operators/Types/user/community/SnapToMultiple.cs
```diff
@@ -0,0 +1,38 @@
...
+        public SnapToMultiple()
+        {
+            Result.UpdateAction = Update;
+        }
+
+        private void Update(EvaluationContext context)
+        {
+            var v = Value.GetValue(context);
+            var mod = Mod.GetValue(context);
+            if (mod == 0){
```
ylmrx, line 21: this part is not needed, in this case the "end test" (second field) of the `for` statement will trigger before the first run, and the loop content won't be executed.

Operators/Types/user/community/SnapToMultiple.cs
```diff
@@ -0,0 +1,38 @@
...
+                return;
+			}
```
ylmrx, line 23: keep indentation consistent for easier code maintenance (your ide can do that for you)

Operators/Types/user/community/SnapToMultiple.cs
```diff
@@ -0,0 +1,38 @@
...
+            for(int goodvalue=mod;goodvalue>0;goodvalue--){
```
ylmrx, line 24: this could be optimized starting from 0 until `iterator==mod`, as we want the minimal value, and the `mod` value can be huge.

also `goodvalue` is not a very good variable name.

Operators/Types/user/community/SnapToMultiple.cs
```diff
@@ -0,0 +1,38 @@
...
+				if(v%goodvalue == 0){
+					Result.Value = goodvalue;
+					return;
+				}
+			}
```
ylmrx, line 29: ditto indent.

ylmrx:
Thanks for this PR ! Thus it's working fine, there's a few cosmetic and optimization issues you can fix 👍

ylmrx:
Regarding the `PlayAtlas` operator itself, there's a bunch of unbound operator you can dispose, and the `ClampInt` ops bound to FrameCount[X/Y] can be removed and replaced with the correct parameter settings (cf. screenshot, set "Min" to 1 in your case).
image

anosci:
I went ahead and made all of the changes you mentioned, except for the loop optimization. I think what you proposed would always return the lowest possible modulo, whereas I'm looking for the next lowest modulo (at, or below, the given input for Mod).

Also, my "IDE" is notepad++, so it might not actually take care of the indentation for me. I'm not sure. 😅
I tried to make it consistent, at least.

pixtur:
Thanks for this awesome contribution!
And thanks to @ylmrx  for going ahead and reviewing. This really helps a lot!

PR #437, opened 2024-05-22
Add [PlayImageSequence] op

anosci:
This adds an op that's an alternative to [PlayVideo]. it uses a series of images in a folder directly, displaying each one sequentially.


pixtur:
Have you seen [ImageSequenceClip] ?

anosci:
I somehow completely missed [ImageSequenceClip]... 

Looking at it, that's a far less featured version of my [PlayImageSequence], and it outputs directly to a <Command>. I've found the ability to output as <Texture2D> to be far more useful. It's quite trivial to make your own timeclip from there.

My op also includes a few QOL features that I mostly made for myself (loop ranges and such), though whether those are *necessary* is debatable.

PR #438, opened 2024-05-22
Feature/pixelate op

newemka:
A "pixel perfect" img fx operator. Its TileAmount parameter allows to set the exact amount of tiles per axis 
 
image


pixtur:
Nice! Although I personally would have called the parameter "Divisions" :-)

PR #439, opened 2024-05-24
Feature/extrude curve compatibility

newemka:
[ac05086](https://github.com/tooll3/t3/commit/ac05086e01dde920bb49497673e60ace57a3e0be) : Fixed [ExtrudeCurves] for some users but broke it for others. 
I've added a boolean parameter allowing to switch between both versions of the HLSL file.
ExtrudeCurves01
ExtrudeCurves02


pixtur:
Thanks. I'm still puzzled that this compute shader performs different on different hardware.

PR #440, opened 2024-05-28
[BoxGradient]

newemka:
sdBox version of [RadialGradient]
This is an implemetation of the signed distance field "Box - Exact" desbribed  in the "2D distance functions" article on Inigo Quilez's website.

image


pixtur:
Do you think it would be possible to merge this with NGon? 
I believe that should be a minor change and would make that op even more powerful.

newemka:
> Do you think it would be possible to merge this with NGon? I believe that should be a minor change and would make that op even more powerful.

True, adding the gradient parameter to Ngon will make it more powerful.

But the shape of NGon is driven by a Radius whereas SdBox size uses width and height. 
Adding a stretch parameter to NGon allows to generate rectangles but the gradient gets stretched.

Then I'd like to keep [BoxGradient] but use sdRoundedBox instead of sdBox so it'll become [RoundedBoxGradient]
https://iquilezles.org/articles/distfunctions2d/

newemka:
image

newemka:
image

pixtur:
Thanks for the contribution

PR #442, opened 2024-05-29
Feature/spring vec2 and vec3

newemka:
Since I enjoy a lot the [Spring] operator, I realized that we need [SpringVec2] and [SpringVec3] as well.


pixtur:
Awesome!

PR #450, opened 2024-07-05
Fix TransformPoints' Rotation and Stretch attributes 

newemka:
Transformation matrix is tricky.
We extract the scale from each row of the TransformMatrix by calculating the length of each row vector.
Then we remove the scale from the matrix to get only rotation.


pixtur:
Nice!

PR #453, opened 2024-07-09
Extend MidiDataRecording.cs

ScreenDream:
Implement Pitchbend and ChannelPressure


pixtur:
Great! Tooll's midi-features become more and more elaborate!

ScreenDream:
Yes! That's the goal :-)

PR #454, opened 2024-07-13
Ndi Output + Screen Capture

mrvux:
As said in the title


newemka:
1. [ScreenCapture] is working very well. 

2. I would set [NdiOuptut]'s FrameRate setting default value to 60. 
R8G8B8A8_Typeless is not working, the description should be replaced by: 

NDI live video output. R8G8B8A8_UNorm texture format is required for output. You can adjust the render format by using a [ConvertFormat]

mrvux:
I would set [NdiOuput]'s FrameRate setting default value to 60.
R8G8B8A8_Typeless is not working, the description should be replaced by

Nobody will use typeless unless they a decent power user so we fine with that, that would need a stream to explain this xD
fell free to accept the PR and change the default

pixtur:
This PR has been merged manually. I'm not sure, why it's not marked as merged on github.

PR #455, opened 2024-07-23
Features/sharpen filter

ScreenDream:
Add a sharpen filter adapted from cables.gl and two example files.


pixtur:
NICE!

pixtur:
I noticed, that you also committed `ReactionDiffusionExample_ddf3077b-6273-4023-88e5-2948312e012b.t3ui`

On dev this will lead to a crash, because the respective .t3 and .cs files are missing.

I fixed the exception and print a warning instead:

image

pixtur:
It actually would be nice to add the ReactionDiffusion example.

ScreenDream:
Hm - weird, I seem to have missed something there then.
Let me try again...

PR #459, opened 2024-07-30
Completing the Gamepad operator and Example

newemka:
By following the first part of gamepad operator, I've added the output for the remaining buttons.

I'm testing with a Xbox one controller, one issue remains, the X button is not working while pressing the Y button turns Y and X bools to true. @mrvux


mrvux:
In GamepadInput,.cs line 31 :
 this.X = y;
should be 
 this.X = x;

newemka:
> this.X = y;

Thanks a lot!

pixtur:
This PR was merged manually

PR #461, opened 2024-08-11
Add [RoundValues] op

Malzbier:
[RoundValues] allows to "round" float arrays values via Round , Floor or Ceiling


Malzbier:
Added Example Setup , Let me know if there is some other documentation / example thing missing,
image

PR #487, opened 2024-11-04
Easings float operator

newemka:
An operator similar to [Spring] and [Damp]
We can select the easing function and set the duration.

This operator creates transitions between changing input values, perfect for animating any step change.
Note: The input value should evolve in steps (e.g., sudden changes) to fully utilize the easing effect.

Easing functions source: [https://easings.net/](https://easings.net/)


domportera:
love this and long overdue 🙏 - i'd definitely want the type selection to be Enum-based though
If you can, pls define the enum in a separate .cs file to avoid issues when duplicating operators

It might also make things simpler on the user end to remove the Strength slot if we can live without it? since it only affects a couple of them, it would confuse me

newemka:
Thanks a lot for your feedback.
I will start working on the separate .cs file now.  
It's true that the Strength is not necessary since for more granular controls we have the Spring operators and they allow to reach the same kind of animation.

domportera:
oh, i didnt mean separate the easing functions into a separate file! But I would _happily_ take that change bc these functions could be useful for the editor >:) or the core math library! hell, the whole functiion `float East(float val, EasingType type)` could be a really nice to have as an editor/core math function. But we can pull that out when these ops get merged into main

what I meant was to make an enum instead of a `switch` statement on an integer (which you've done!) and have *that* in a separate file, so the input Ease would be
`InputSlot<EasingType> Ease = new();`

Seems like you're already in the process of doing so, just wanted to clarify bc i feel bad if I encouraged you to do more work than you needed to 😅

I also say this not knowing how well the editor supports just adding an enum slot like that though... 🤷‍♂️

newemka:
I think the way I did it is okay. I have the enum and functions in [EasingFunctions.cs] So I can use them in any operator like Easings or EasingsVec3. Creating EasingsVec2 will be straight forward. 

About `InputSlot<EasingType> Ease = new();` , I have no idea how I could do that, I'm just following how it's done in the other operators, with an interger. 

```
[Input(Guid = "9ef45ff9-9d59-44b7-b03d-37a1be40a776", MappedType = typeof(EasingFunctions.EasingType))]
        public readonly InputSlot<int> Ease = new();
```

Right now the only way to see the dropdown menu with the enums in the parameters window of an operator is to add , `MappedType = typeof(EasingFunctions.EasingType)` and the enums to the cs file of an operator. 

Thus, duplicating such operator within the editor will give a int and won't show a dropdown menu.

newemka:
This how the parameter window looks like 
image

PR #490, opened 2024-11-25
Update nuget packages in order to avoid vulnerabilities

newemka:
When opening the solution, Visual Studio shows a warning message about nuget packages vulnerabilities. 
I've used the nuget packages manager and selected the packages versions who are not marqued as vulnerable.


pixtur:
Thanks! That's a great update.
We can use this data to also update main, once is a bit more stable.

newemka:
When opening Main on Visual Studio, I don't get these vulnerability warnings. I think it's beacause Main is using a more recent version of .Net

PR #493, opened 2024-12-22
Fix DrawBillboards AtlasMode

newemka:
Scatter and UseW atlas modes are now using every images from a square atlas map.
Big thanks to @sioda-ie  for submiting this fix on Discord.


pixtur:
Thanks for that quick fix!

PR #499, opened 2025-01-13
Alternative version of KeyboardInput 

newemka:
This version gives access to the number keys from the top row of the keyboard and the numpad. It outputs the corresponding int value of the pressed key.

I'm using it to trigger the appearance of overlays while performing live.

A problem remains, I'm often using the numpad to enter values in the parameter window, then it triggers my overlays.
How can I avoid the register the pressed number keys while editing a setting value?


newemka:
Closed because it has been implemented in TiXL 4.0.3

PR #500, opened 2025-01-16
BoxGradient Fixed on Main

newemka:
Includes path fixed.
Rotation input parameter was connected twice to _multiImageFxSetupStatic.

For some reason I had to restart T3 before seeing the op working properly.


pixtur:
Thanks for the quick fix!

PR #502, opened 2025-01-20
Feat/equirectangle improvement

newemka:
Same as previous PR for main 
Using CameraWithRotation and SamplerStateOp's AddressU and V to Clamp


pixtur:
Thanks!

PR #503, opened 2025-01-20
Project creation fix.

banidu-512:
Project property updated to net9.0.

### DIFF
--- a/Editor/Compilation/CsProjectFile.Defaults.cs
+++ b/Editor/Compilation/CsProjectFile.Defaults.cs
@@ -12,7 +12,7 @@ internal sealed partial class CsProjectFile
     private static readonly FrozenDictionary<PropertyType, string> DefaultProperties =
         new[]
                 {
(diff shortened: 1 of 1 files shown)


domportera:
hey thank you for this PR! unfortunately (and fortunately!) I made some changes today that should make this issue a thing of the past, but it renders this PR unmergeable /:

banidu-512:
> hey thank you for this PR! unfortunately (and fortunately!) I made some changes today that should make this issue a thing of the past, but it renders this PR unmergeable /:

Hmm interesting that it is not mergeable. I just changed the version number :D in this way that you changed it im going to close it.

domportera:
yes I made it so that .NET versions arent mentioned in this area at all - now the `RuntimeAssemblies` class holds a single reference to the .NET runtime used to compile Tooll, and the projects get updated to that! In the future i'll probably replace this literal string "9.0" to something actually retrieved at runtime, but ya

PR #510, opened 2025-01-28
Fix audio input for small buffer audio devices

newemka:
Same as for Dev, small buffer audio devices such as USB microphones are working now.


pixtur:
Thanks!

PR #511, opened 2025-02-01
TexCoord2 support for [DrawMesh] & [SetMaterial] 

newemka:
Since I've updated the PBR struct for [DisplaceMeshVAT], a mesh can have a second set of texture coordinates. 
 

https://github.com/user-attachments/assets/e61e3357-db5f-4017-9f5c-20bb4f0ff6f7

Todo :

- [x] support for shared Blend-function.hlsl
- [ ] TexCoord switches for the different textures
- [ ] trying to not over bloat [SetMaterial]

### DIFF
--- a/Core/Rendering/Material/PbrMaterial.cs
+++ b/Core/Rendering/Material/PbrMaterial.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Numerics;
 using System.Runtime.InteropServices;
 using SharpDX.Direct3D11;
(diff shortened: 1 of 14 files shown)


pixtur:
Hi @newemka,

I would have thought that many more than 14 files would require to be updated for that.
Shouldn't this affect all mesh operators?

newemka:
You're right, since the PBR struct update , all mesh operators have their stride updated as well. (https://github.com/tooll3/t3/pull/479)
Right now the state of this PR is just what I needed for a project I'm working on.  It allows me to see to make some experiments and see how to implement the use of the TexCoord2 for more operators. 
I will update more mesh opertors soon.

pixtur:
Okay. No worries. But it would probably be good to pull this over to main

PR #513, opened 2025-02-13
Easing Operators : Ease, EaseVec2 and EaseVec3 

newemka:
In order to avoid code duplication I've moved the switch to EasingFunctions.cs


pixtur:
Nice!

PR #514, opened 2025-02-21
Feature main/noises

newemka:
FractalNoise and WorleyNoise can blended together, the same offset applied to both won't result with a kind of parallax.


pixtur:
Awesome!

PR #518, opened 2025-03-15
Fix and improve NgonGradient

newemka:
sdRegularPolygon function is replacing sdNgon. 
I've managed to bring back the Curvature parameter in the SdRegularPolygon function 
I didn't find how to add the updated presets.


pixtur:
Awesome!

PR #520, opened 2025-03-20
Pbr vertex Texcoord 2 

newemka:
I've updated the PBR vertex struct and the stride for most of the Mesh modifier operators


pixtur:
It was merged on main. I'm not sure, why github is showing it as unmerged here...

PR #522, opened 2025-03-24
fix and improve extrude curves

newemka:
I broke ExtrudeCurves with PR #520 
Here is the fix, I made sure it generates proper UV maps for TexCoord and TexCoord2 
extrudeCurves001
extrudeCurves00


newemka:
I close this PR, I'll make a new one without conflict.

PR #527, opened 2025-03-27
Add JumpFloodFill operator

0xNULLderef:
Unsure if lib.image.generate.misc is the best place to put it, but this should be at least a good starting point for an implementation in tooll.

Depends on [the Log pull request](https://github.com/tooll3/t3/pull/526) as that isn't possible to implement standalone.


pixtur:
Ohhhh. That's so awesome. That is a really great contribution.

PR #528, opened 2025-03-28
TexCoord2 support for [DrawMesh] & [SetMaterial]

newemka:
See #511


pixtur:
Although I think that this is a great feature to have, I think we should look for a more generic solution probably with help of the shader graph nodes.

Issue #532, opened 2025-04-07
Combine "Render Video Window" and "Render Sequence Window" into "Render Window"


newemka:
### Problem: How to render an image sequence? 

Users often wonder how to render an image sequence, 

### Solution

A single Render Window where we can choose to render to a video file or to an image sequence

Here are screenshots of how the Render window could look:

![Image](https://github.com/user-attachments/assets/886d40e6-2a95-4cab-b392-a5a450a4e4ce)

![Image](https://github.com/user-attachments/assets/5532791b-b3c6-4818-b3a5-9b719c056bf1)

pixtur:
What is the status if this? I remember it looks very promising at the meet-up...

newemka:
Image Sequece: The File name input field is being ignored

newemka:
With the last commit I made the image sequence naming is working. 
But something is bothering me, I have BaseRenderWindow.cs for the common parameters and RenderWindow.cs who takes care of video and image sequence. 
In order to be future proof I wonder if it would make more sense to split RenderWindow.cs into two files.

newemka:
#534 is ready for review

A problem remains I think OutPutWindow.cs is not properly telling when its content can be rendered as video or image file. 
This is also a problem for making screenshots.

pixtur:
PR is merged. Excellent progress!

PR #534, opened 2025-04-07
POC single Render window for video and image sequence

newemka:
This is a rough proof of concept to replace "Render Video Window" and "Render Sequence Window" by a "Render Window", see this feature request : #532 

What's missing? 

Image Sequence's Prefix is not working.
Tooltips : to help the user understand where the files will be written by default (inside the Project's folder) 
Maybe adding Reset button like in the setting window for the folder path


pixtur:
Super cool improvement! Works like a charm!

PR #535, opened 2025-04-09
ExtrudeCurves update

newemka:
If UVsDirection is false: TexCoord -> Horizontal | TexCoord2 -> Vertical
image

If UVsDirection is true: TexCoord -> Vertical | TexCoord2 -> Horizontal
image


pixtur:
Good catch!

PR #537, opened 2025-04-10
A quick update of [VisualizePoints]

newemka:
I've replaced the obsolete [_OffsetPoints] by [TransformPoints] and removed duplicate operators so it looks similar as before.


pixtur:
Awesome!

PR #539, opened 2025-04-14
CI System

Stefterv:
I got curious today to see if I could add the Windows signing that we (@pixtur) talked about and I noticed there is no CI system. 
Since I have recently done this for Processing I figured it could be helpful here too.

This PR adds the following:
- Build and Package on main to create nightly builds
- Add a Build action to pull requests to check if pull requests provide a functional build at least.
- A release action that will build and package Tooll and upload the release binary to the created release.
   - This will also take the value of the tag associated with the release as the version name, to prevent the manual step of modifying the version for each release

### DIFF
--- a/.github/actions/build/action.yml
+++ b/.github/actions/build/action.yml
@@ -1,0 +1,11 @@
+name: "Build"
+description: "Run common build steps"
+runs:
+  using: "composite"
+  steps:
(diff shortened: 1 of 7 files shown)


pixtur:
Holy crap! Thank you!

PR #544, opened 2025-04-22
Fixed BeatTimingPlayback not updating FrameCount

0xNULLderef:
All raymarching ops should now properly update their parameters.
The bug was caused by the check
`if (_lastUpdateFrame == Playback.FrameCount)`
in `ShaderGraphNode.Update` always failing since the FrameCount stayed at 0, and was introduced in [this commit](https://github.com/tooll3/t3/commit/791887b39d619355d4876f906ccde83a64dc3790).


pixtur:
Excellent catch! Thanks for the fix!

PR #546, opened 2025-04-26
TriangularPrismField : Split RadiusLength into Radius and Length

newemka:
For better user experience.
Sadly this 3d SDF can't be rounded 
image


pixtur:
Thanks!

PR #547, opened 2025-04-29
fix github action name

NikkyAI:
small typo we noticed.. this fixes the labelling of actions in github: 

screenshot


pixtur:
Nice. Good catch.

PR #555, opened 2025-05-05
Add Polar Mapping to RaymarchField

newemka:
image


pixtur:
Thanks for this! But I think that the proper solution should be a [PolarMapUVSpace] operator.
I can look it this.

PR #563, opened 2025-05-10
Improvement: Read commit hash from editor product version field

0xNULLderef:
Removes git dependency on running system.

The original would've had an undefined behavior when running on a system without the git client installed or if the executable had been moved outside the git repository tree.

Visually it is identical to the original.

image


domportera:
this looks great! excellent thinking that would have otherwise gone unnoticed (until it's a problem for someone)

just would request in the future that the formatting changes be removed for readability to prevent unnecessary merge conflicts that aren't related to the meat of the PR. trust me when I say that I have done it myself many times, just ask pixtur 😂 thank you!!

PR #564, opened 2025-05-10
Improvement: chagned behavior of BYPASS button

0xNULLderef:
The button is grayed out when an operator cannot be bypassed. A tooltip is shown a when hovered to inform the user it isn't possible.

Screenshots of:
- normal bypassable operator
TiXL_7cpijkP6XZ

- non bypassable operator
TiXL_9e4nU3F5XN


pixtur:
Wow. Super cool!

PR #565, opened 2025-05-11
Update Stride Vertices for RepeatMeshAtPoints

sense-amr:
Use Stride out from Vertices to NewVertexBuffer

https://github.com/user-attachments/assets/238934dd-580b-41d7-8d32-e2a56d1fde31


pixtur:
Nice. Good catch.

PR #567, opened 2025-05-13
Feature main/exit ask before quit

newemka:
It works with Exit from TiXL menu and with the X icon at the top right of the application window.
image


newemka:
TiXL_JNWCRjbZ1c

pixtur:
Thanks a lot! That's really awesome.

PR #573, opened 2025-05-18
Fix ScreenShots feature

newemka:
## Problem: 
No matter the selected operator `GetCurrentTexture()`  is never null so the Screenshot button is always visible. 

While in Debug mode, clicking the Screenshot button was creating a Screenshot folder in _Editor\bin\Debug\net9.0-windows_

If a _Command_ or a _Texture2D_ operator is selected we get a png file with the content of the Output View.

If an other kinf of operator is selected we get a png file with the last render (from a Command or Texture2D) of the Output View. 

## Solution:
Same logic as for the _"Start Render"_ button in the Render To File window

But here we allow to save screenshots for _Texture2D_ and _Command_ operators.

If it doesn't exist a _Screenshots_  folder is created in the opened project's folder and the png image is written inside.  

I also added the Screenshots folder to the list of excluded directories in AutoBackup.cs

### DIFF
--- a/Editor/Gui/AutoBackup/AutoBackup.cs
+++ b/Editor/Gui/AutoBackup/AutoBackup.cs
@@ -54,7 +54,7 @@ internal static class AutoBackup
 
         var zipFilePath = Path.Join(BackupDirectory, $"#{index:D5}-{DateTime.Now:yyyy_MM_dd-HH_mm_ss_fff}.zip");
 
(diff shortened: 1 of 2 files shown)


pixtur:
Oh good catch!

sense-amr:
There should be documentation added as to where to NOW find the screen shots .. 
and it essentially makes the existing Editor\bin\Release\net9.0-windows\Screenshots redundant .. 
It should be stated in documentation that the location of Screenshots is now .. 

...Documents\TiXL\projectname\Screenshots

PR #577, opened 2025-05-22
Editor will ignore the mouse if it's on the Output Window

newemka:
## Problem: 
Editor UI elements with is hovered behaviors are triggered by the mouse over the Output Window.
## Solution: 
Make ImGui ignore the Output window.

Fix UI flickering or unexpected highlights in the Editor due to false hover states when the mouse is hovering the Output Window.


newemka:
Thanks, I'm pretty proud of this one :)

PR #593, opened 2025-06-03
Add dot product and cross product ops

Martysh12:
Somehow, there are no dot product or cross products ops. This PR adds those.

This PR adds 4 new ops:
- `DotVec2`
- `DotVec3`
- `DotVec4`
- `CrossVec3`

image

These are quite essential when doing 3D graphics.


pixtur:
Thanks a lot!

PR #598, opened 2025-06-04
Arc Connection Segments Setting

newemka:
Adds a new Magnetic Graph setting to let user change the segment amount of the arc connection. 
image

Also a small fix to adjust the node selection line rounded corner match to the node.
image


pixtur:
Nice!

PR #601, opened 2025-06-10
HueShift op

newemka:
Adjust Color is doing too much things, here a a simplified version that concentrates on hue shifting, saturation and exposure.


pixtur:
I agree with [AdjustColor] doing to much. But [HueShift] implies doing to little (or at least less than the descriptions suggests).

Why not call it HSB?

newemka:
Yes, good point! I was thinking the same about the name, but B is for Brightness.
Here the difference between Exposure and Brightness is that Exposure doesn't turn black pixels into white.
Thus, to be accurate it needs to be called HSE.

PR #602, opened 2025-06-10
TorusSDF: split Size parameter into Radius and Thickness just like TorusMesh

newemka:
image

Examples which are using TorusSDF are updated.


pixtur:
Ahhh... Damn That will break so many of my examples. But it's definitely better!

newemka:
I only found TwistFieldExample and CombineSDFExample and fixed them

PR #603, opened 2025-06-12
fixes the dotnet sdk in the installer

NikkyAI:
use correct url for downloading the dotnet sdk
set name of created artifact in nightly build


pixtur:
Good catch! Thank you

PR #605, opened 2025-06-13
AttributeFromImageChannels Rotation factors fix

newemka:
Only Rotate_X was working 
There was an issue between degrees and radians


pixtur:
Oh. Good catch. Thanks for fixing.

PR #614, opened 2025-06-21
Rotation gizmo

newemka:
This is the base of the Rotation Gizmo

TiXL_ssupQ4kbjG


pixtur:
I think this is an incredible start, but the euler angles need to be handled differently (probable not stay locked 90 deg).

PR #692, opened 2025-06-24
GC On rendering Image

mrvux:
Make Image render much more GC friendly, convert the using clause (which does some allocations for type checks), to a basic null check


pixtur:
Thanks!

PR #693, opened 2025-06-26
Quick update and Example: [DrawMeshChunksAtPoints] 

newemka:
DrawChunksAtPoints.hlsl needed to be updated to the new Point struct.

(There is still room for improvement, I'm thinking about F1 and F2 support) 

I added a quick example.


pixtur:
Excellent. Thank you!

PR #705, opened 2025-07-10
Quick fix of SelectPointsExample

newemka:
It needed to be updated to the new point struct. 
image


pixtur:
Awesome! Let's add a test for this.

PR #709, opened 2025-07-13
Fix SplitMeshVertices

newemka:
It was producing a ridiculous amount of vertices and could cause crashes. 
Now the amount of vertices is 3 per face.


pixtur:
Good catch!

PR #711, opened 2025-07-14
Add utility to create Dynamic constant buffer (takes care of 16 align)

mrvux:
Add utility to write to Constant buffer in GC free manner


pixtur:
I have to review this again, because it lead to UI artifacts on more complex graphs.

Issue #717, opened 2025-07-27
v4.0.5.0 - Order project hub by recency


edave64:
### Is your proposal related to a problem?

As I understand the code, the current ordering of the project hub is just whatever EnumerateDirectories returns. That order is not guaranteed to mean anything. And new projects always appear at the end. That is almost certainly not where the user expects it, since they probably want to work at that new project now.

### Describe the solution you'd like

I think the default ordering should be when the project was last updated or maybe when it was last opened. With new projects always appearing at the top of the list.

Sorting by name is still useful for finding old files, so maybe there could be a selector for how you want to sort it. But then the name sorting should be consistent, including new projects.

PR #718, opened 2025-07-28
Sort project hub by descending creation date

edave64:
This addresses https://github.com/tixl3d/tixl/issues/717.

There are two commits here:
1) Add a `CreatedAt` field in `EditableSymbolProject` and use it to sort the Project Hub. The data for this field is read from the already existing fileinfo handle, or set as the current time when creating a new project.
2) Add a cache to the `AllProjects` projects list, so it doesn't have to be filtered and sorted every time the Project Hub refreshes. The cache is invalidated every time a `EditableSymbolProject` is created or disposed.

Since this is my first code contribution to this project, and don't know any ImGUI projects, I don't know if the cache is overzealous or if caching should be done some other way. Commit 1 can work without it, and I'm open to change request.

### DIFF
--- a/Editor/Compilation/CsProjectFile.cs
+++ b/Editor/Compilation/CsProjectFile.cs
@@ -55,9 +55,15 @@ internal sealed class CsProjectFile
     /// </summary>
     private string TargetFramework => _projectRootElement.GetOrAddProperty(PropertyType.TargetFramework, ProjectXml.TargetFramework);
 
-    private CsProjectFile(ProjectRootElement projectRootElement)
+    public DateTime CreatedAt => _fileInfo.CreationTimeUtc;
+    public DateTime ModifiedAt => _fileInfo.LastWriteTimeUtc;
+
+    private CsProjectFile(ProjectRootElement projectRootElement) : this(projectRootElement, new FileInfo(projectRootElement.FullPath)) { }
+
+    private CsProjectFile(ProjectRootElement projectRootElement, FileInfo fileInfo)
     {
         _projectRootElement = projectRootElement;
+        _fileInfo = fileInfo;
 
         var targetFramework = TargetFramework;
         
@@ -199,9 +205,10 @@ internal sealed class CsProjectFile
     /// Does not actually handle any assemblies or type loading - it's just a way to load the xml file.
     /// </summary>
     /// <returns>True if successful</returns>
-    public static bool TryLoad(string filePath, out CsProjectLoadInfo loadInfo)
+    public static bool TryLoad(FileInfo fileInfo, out CsProjectLoadInfo loadInfo)
     {
         bool success;
+        var filePath = fileInfo.FullName;
         try
         {
             var fileContents = ProjectRootElement.Open(filePath);
@@ -212,7 +219,7 @@ internal sealed class CsProjectFile
             }
             else
             {
-                loadInfo = new CsProjectLoadInfo(new CsProjectFile(fileContents), null);
+                loadInfo = new CsProjectLoadInfo(new CsProjectFile(fileContents, fileInfo), null);
                 success = true;
             }
         }
@@ -283,6 +290,7 @@ internal sealed class CsProjectFile
         try
         {
             _projectRootElement.Save();
+            _fileInfo.Refresh();
         }
         catch (Exception e)
         {
@@ -298,6 +306,24 @@ internal sealed class CsProjectFile
     public bool TryCompileRelease(bool nugetRestore)
     {
         return Compiler.TryCompile(this, PlayerBuildMode, nugetRestore);
+    }
+
+    /// <summary>
+    /// Updates the last modified date of the project file by re-saving the project XML.
+    /// </summary>
+    public void UpdateLastModifiedDate()
+    {
+        // we can set the last modified date by re-saving the project xml
+        using var textWriter = new StreamWriter(_projectRootElement.FullPath);
+        try
+        {
+            _projectRootElement.Save(textWriter);
+            _fileInfo.Refresh();
+        }
+        catch (Exception e)
+        {
+            Log.Error($"{Name}: Failed to save project file {_projectRootElement.FullPath}: {e}");
+        }
     }
 
     // todo- use Microsoft.Build.Construction and Microsoft.Build.Evaluation
@@ -375,6 +401,7 @@ internal sealed class CsProjectFile
     private readonly string _releaseRootDirectory;
     private readonly string _debugRootDirectory;
     private readonly ProjectRootElement _projectRootElement;
+    private readonly FileInfo _fileInfo;
 
 
     public void IncrementBuildNumber(int amount)
--- a/Editor/Compilation/ProjectSetup.Startup.cs
+++ b/Editor/Compilation/ProjectSetup.Startup.cs
@@ -142,7 +142,7 @@ internal static partial class ProjectSetup
         {
             if (projectInfo is { csProjFile: not null, success: true })
             {
-                var project = new EditableSymbolProject(projectInfo.csProjFile);
+                var project = new EditableSymbolProject(projectInfo.csProjFile, projectInfo.fileInfo.CreationTimeUtc);
                 AddToLoadedPackages(project);
             }
             else
--- a/Editor/Compilation/ProjectSetup.Startup.cs
+++ b/Editor/Compilation/ProjectSetup.Startup.cs
@@ -117,7 +117,7 @@ internal static partial class ProjectSetup
                       .AsParallel()
                       .Select(fileInfo =>
                               {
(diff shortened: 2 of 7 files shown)


edave64:
Of course, this doesn't sort by last used, which I'd really like and proposed in the linked issue. But I'm not sure how to do that without larger changes and additional loads, which I wanted to avoid since I'm not all that deep into the code-base.

domportera:
this looks good, but involved more logic-specific code than I think is necessary -

Each `CsProjectFile` contains its project XML inside of it - you could simply expose the creation/modification dates via a public property or get method (and if necessary, expose that within `EditableSymbolProject`)

domportera:
and to answer your question - I think your enumerable cache is quite elegant. I would probably add an invalidation (i.e. `cache = null`) to the Dispose method too to be future-safe if this doesn't generate exceptions when the app exits

edave64:
@domportera There already is an invalidation in the Dispose method.

I'll take a look at the project XML and how it works.

domportera:
oh good, thanks!

edave64:
@domportera I might have misunderstood, are you suggesting just to add these fields to the `CsProjectFile` class or to actually save it in the `.csproj` file?

I could do the former, but it would be a slightly awkward construction, since, as far as I see, `CsProjectFile` has no reference to the actual file. It has `projectRootElement.LastWriteTimeWhenRead`, unfortunately it seems the `.csproj` file is only modified if new Symbols are added to the project. So it works as neither as a reliable creation nor last modification.
So I would either have to drag the creation time trough to its constructor, creating the same issue just with a different class and across more methods, or I make it a public setter and set it after construction from the outside, which would be a strange thing to do. Or add a new Filesystem call to the `CsProjectFile` constructor to re-fetch information that already exists further down the callstack.

As for a real modification time, I think I would have to read that from the main symbol file. It gets modified whenever the project is saved, but unfortunately isn't available at construction time of `EditableSymbolProject`. So it have to wait for the main symbol to load (The HomeSymbol, it think?) and then check its modification time. That's even further removed from the `CsProjectFile`, and just semantically wouldn't make sense to put there.

Would it be OK to add a (blocking) filesystem call to the adding of symbols to a `EditableSymbolProject`? I don't currently see any other way to get a good modification date otherwise.

domportera:
> @domportera I might have misunderstood, are you suggesting just to add these fields to the `CsProjectFile` class or to actually save it in the `.csproj` file?
> 
> I could do the former, but it would be a slightly awkward construction, since, as far as I see, `CsProjectFile` has no reference to the actual file. It has `projectRootElement.LastWriteTimeWhenRead`, unfortunately it seems the `.csproj` file is only modified if new Symbols are added to the project. So it works as neither as a reliable creation nor last modification. So I would either have to drag the creation time trough to its constructor, creating the same issue just with a different class and across more methods, or I make it a public setter and set it after construction from the outside, which would be a strange thing to do. Or add a new Filesystem call to the `CsProjectFile` constructor to re-fetch information that already exists further down the callstack.
> 
> As for a real modification time, I think I would have to read that from the main symbol file. It gets modified whenever the project is saved, but unfortunately isn't available at construction time of `EditableSymbolProject`. So it have to wait for the main symbol to load (The HomeSymbol, it think?) and then check its modification time. That's even further removed from the `CsProjectFile`, and just semantically wouldn't make sense to put there.
> 
> Would it be OK to add a (blocking) filesystem call to the adding of symbols to a `EditableSymbolProject`? I don't currently see any other way to get a good modification date otherwise.

so i was just referring to creation date, as you are indeed correct that the only time the csproj file's last-write attribute would be changed is when the project recompiles (as we update version numbers and whatnot when recompiling), which is not a great metric for last-use.

that being said, modification of the "home" symbol may also not be the best metric for this (but is definitely better), as lots of work could be done in other symbols that don't affect the "home" symbol. this to me seems a bit harder, as it has a potential consequence of 1. slowing down the loading process if we need to check file timestamps while we load them, or 2. introducing additional logic that seems like itll get more and more complicated the more the feature is improved

that's basically why I suggested what I did - the idea is that there is no date information passed into the project or into the csproj class. the csproj class itself has a reference to its project xml baked in, and so a couple of "get" properties could be made, or even a csproj attribute could be changed and saved such that its modification time is changed - in this way, the most complex case is actually very simple. here's what I mean (pls excuse the pseudo-code): 

```cs
internal partial class CsProjectFile
{
	// we can use a FileInfo object instead of the _projectXml since this class has a reference to its file path
	// we can likely cache a FileInfo in the constructor of this class
	public DateTime LastModifiedDate => _fileInfo.LastModifiedDate; 
	public DateTime CreationDate => _fileInfo.CreationDate;
}
```

the most complex case, where we can force the `LastModifiedDate` to update when the project doesn't necessary compile between work sessions, is to add something like this as well:

```cs
internal partial class CsProjectFile
{
	// ^ same as above ^

	public void UpdateLastModifiedDate()
	{
		// we can set the last modified date by re-saving the project xml
		using var textWriter = new TextWriter(_fileInfo.Path);
		try
		{
			_projectXml.Save(textWriter);
			_fileInfo.Refresh();
		}
		catch (Exception e)
		{
			// ..... error log etc
		}
	} 
}

```

we'd need to use [this particular overload](https://learn.microsoft.com/en-us/dotnet/api/microsoft.build.construction.projectrootelement.save?view=msbuild-17-netcore#microsoft-build-construction-projectrootelement-save(system-io-textwriter)) due to the fact that it allows us to save when the data hasn't changed.

When an `EditableSymbolProject` is saved, we can simply add a call to `this.CsProjectFile.UpdateLastModifiedDate()`, and we'll have what we want with as little pain as possible

edave64:
@domportera I've added some code to track creation and modification date through the `FileInfo` in `CsProjectFile`.

Is this what you had in mind?

For some reason I didn't notice that `FileInfo` also has `LastWriteTimeUtc`. 😅

The new code actually sorts by modification date, so it also needs to a few more cache invalidations, but now no extra filesystem access happens on initial load.

I also noticed that, as a side-effect, the "recent projects" list in the menu also sorts by last modified, which I see as a win :P

domportera:
this looks wonderfully done! i'd love to be irresponsible and just merge it without testing it bc it looks good, so I think I will. if anything is wrong with this feature, it is entirely my fault and i am to blame. thanks a bunch for your contribution!

newemka:
> this looks wonderfully done! i'd love to be irresponsible and just merge it without testing it bc it looks good, so I think I will. if anything is wrong with this feature, it is entirely my fault and i am to blame. thanks a bunch for your contribution!

@domportera 

Since this merge, I'm confused, If I understand well once I save a project it should appear at the top of the projects list. But it's never happening, I don't understand the ordering of the project list.

newemka:
@edave64

domportera:
> > this looks wonderfully done! i'd love to be irresponsible and just merge it without testing it bc it looks good, so I think I will. if anything is wrong with this feature, it is entirely my fault and i am to blame. thanks a bunch for your contribution!
> 
> @domportera
> 
> Since this merge, I'm confused, If I understand well once I save a project it should appear at the top of the projects list. But it's never happening, I don't understand the ordering of the project list.

interesting.. it's probably a missed cache invalidation step `EditableSymbolProject._allProjectsCache`(or something) or something cursed with the save function that conflicts with microsoft's docs (hopefully not that)

as a patch (which i cant do atm bc im at work), I would try doing a `_allProjectsCache = null` at [Editor/UiModel/EditableSymbolProject.FileHandling.cs line 26](https://github.com/tixl3d/tixl/pull/718/files#diff-ad9c6efb6cdc52d3784509db58cc19c4f3a30b5c45a536f32a267f53cd168d31)

edave64:
The projects are sorted by the modification date of the csproj files. 

For projects saved after this PR, that should be the time of the last save.

For projects past saved before the PR, as I understand it, it's the time of last compilation. That's either the time of creation or when you last declared a new symbol in the project, as far as I know.

@newemka Does the ordering of the project list change when you reopen the program?

newemka:
I run Tixl then open a project. 
Then back to the Project Hub. 
Every 3 mins (auto save) the opened project is getting a new place. 
image

newemka:
image

edave64:
That sounds like more than just an invalidation issue. The strange thing is that `OrderBy` is stable. So if `ModifiedAt` weren't there, I would expect it to just sort alphabetically.

Can you log the `ModifiedAt` values per project [here](https://github.com/tixl3d/tixl/pull/718/files#diff-04a7ca48faa6baaac8465ffc0aaab7718a29760bc82d58574023b766ae9b4107R154) and check what's going on with those values?

edave64:
I've tested it and can confirm the bug in the current master.

Autosave seems to update every csproj file, effectively randomizing the sorting. I'm trying to figure out why.

PR #720, opened 2025-07-30
Support filenames with spaces when opening files or folders externally

edave64:
Wrap the given URI passed to "cmd /c start" in quotes, so it can handle spaces inside filenames.

This is especially important with newer tixl versions installing themselves in "C:\Program Files". Currently, they cannot open the folders and files of the projects they ship with.

A screenshot of an error saying 'D:/dev/test' could not be found, with a console in the background showing tixl running in 'D:/dev/test with spaces/'

(In the above example I pressed the "Edit" button on one of the SVG files of "Examples.user.still.there" with tixl running in "D:/dev/test with spaces/")


domportera:
thanks!! great catch

PR #728, opened 2025-08-09
Add [OscOutput] FloatList input.

banidu-512:


pixtur:
I like the idea and the code looks clean and well structured. But it would be great to add more details for upcoming ops. 
I took some time to update a small checklist on https://github.com/tixl3d/tixl/wiki/dev.OperatorConventions 

Especially a single line of documentation is always super useful.

PR #731, opened 2025-08-11
Add [ArtnetInput] node for listening artnet.

banidu-512:


pixtur:
Thanks for that contribution. I think there are some unnecessary allocations in the `Update()` method that might lead to framedrops if used excessively... This should probably be refactored eventually.

PR #732, opened 2025-08-11
Update [ArtnetOutput] with raw networking without library. Update [PointToArtNetLights] Nan correction.

banidu-512:
ArtnetOutput:
-Now raw networking
-Added unicast and artsync. Theres poll functions now. 

PointToArtNetLights:
-added white function
-pan/tilt can be disabled
-colors now clamped and corrected for NaN


pixtur:
Merged in Main but I probably have to review and test this in depth.

PR #733, opened 2025-08-11
Don't update project modification time if no symbols were changed

edave64:
When the automatic save runs, it currently always updates the project csproj, regardless of any modification, creating effectively a random sorting. (Reported here: https://github.com/tixl3d/tixl/pull/718#issuecomment-3175086505)

This fixes that.

@domportera 
@newemka


edave64:
Sorry for the mistake, I guess I never let it run long enough in my tests to trigger the autosave?

domportera:
> Sorry for the mistake, I guess I never let it run long enough in my tests to trigger the autosave?

no worries! it was my fault anyways :) no harm done. thanks for the quick resolution!

newemka:
This is much better :) thanks a lot for the quick fix!

PR #735, opened 2025-08-11
fix BendField operator

NikkyAI:
The op only worked as the last link in the chain before


pixtur:
Awesome! Thanks for that contribution

PR #736, opened 2025-08-13
Fix Trigger OnlyOnDown

newemka:
When OnlyOnDown was true the result was always true until we release the button.


pixtur:
perfect!

PR #741, opened 2025-08-20
Hopefully better BoxSDF uniform scale

newemka:
Please check the code. I think this version is prettier.


pixtur:
I implemented a better solution during the live screen.

PR #749, opened 2025-08-26
add DivideVector2

NikkyAI:
there was no way to do this easily and it is required for calculating texture scaling and the like
tried my best to keep the descriptions updated


pixtur:
Good catch!

PR #751, opened 2025-08-27
Avoid Hovering Preview to show up in the Output Window

newemka:
This is an attempt to fix these problems: 
1. Hovering a Texture2D type connection line is overriding the pinned render in the Output Window (not ideal in VJing context)
2. MagGraph is missing the implementation of the different UserSettings.GraphHoverModes states (set by the TimeControl's ToggleHover button)
3. Hovering an operator output or input slot has the same issues as hovering a connection line

I think I managed to fix 1 and 2. 
I'm still looking for how to apply the same fixes for problem 3.

Edit: ok I understand now... Instead of working on DrawTooltipForSingleOutput method I should have worked on the ToolTipContentDrawer Class

### DIFF
--- a/Editor/Gui/MagGraph/Interaction/ConnectionHovering.cs
+++ b/Editor/Gui/MagGraph/Interaction/ConnectionHovering.cs
@@ -1,4 +1,4 @@
-﻿#nullable enable
+#nullable enable
 using ImGuiNET;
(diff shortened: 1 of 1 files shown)


pixtur:
Okay. I can look into that.

PR #753, opened 2025-08-28
Implementing GraphHoverModes for MagGraph

newemka:
and Sync button size fix


pixtur:
Nice. Good catch, btw.

PR #756, opened 2025-08-30
ReactionDiffusion

newemka:
Add Clamping option to `Sharpen`, it helps to get nice and smooth reaction diffusion. 
Rework of the Reaction Diffusion Example


newemka:
- The original example was working only if the user press play on the timeline. I fixed this by adding a [Once]. 
- Non breaking changes for people using [Sharpen], the Clamping is a new bool parameter.  
- [SharpenExample] remains identical 
image

pixtur:
So cool!

PR #763, opened 2025-09-03
fixes [SetBpm] by using the BpmProvider Instance directly

NikkyAI:
BpmProvider and TapProvider moved to core
interfaces are no longer needed, but i haven't discarded them yet


pixtur:
Nice!

PR #766, opened 2025-09-05
[CustomPixelShader] add inputs for function definitions when using template, examples

ylmrx:
- add some multi-line string input to bring in function snippet (making the usage with Template much more convenient
- add an example showing both usages

### DIFF
--- a/Operators/Lib/image/use/CustomPixelShader.cs
+++ b/Operators/Lib/image/use/CustomPixelShader.cs
@@ -31,6 +31,9 @@ internal sealed class CustomPixelShader : Instance<CustomPixelShader>
     [Input(Guid = "fb8d51fe-b4c2-452a-9e53-b649aed92bd7")]
     public readonly InputSlot<bool> IgnoreTemplate = new();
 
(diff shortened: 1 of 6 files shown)


Operators/Lib/image/use/CustomPixelShader.cs
```diff
@@ -31,6 +31,9 @@ internal sealed class CustomPixelShader : Instance<CustomPixelShader>
     [Input(Guid = "fb8d51fe-b4c2-452a-9e53-b649aed92bd7")]
     public readonly InputSlot<bool> IgnoreTemplate = new();
 
+        [Input(Guid = "ce2d62f5-fa99-4bb2-8781-d5fe44c10909")]
+        public readonly MultiInputSlot<string> AdditionalFunctions = new MultiInputSlot<string>();
```
pixtur, line 35: Oh. Making this a multiInput is very unfortunate, because it's values will no longer be stored by presets. I _really_ recommend to make this a single string input. 

Users can use [CombineStrings] to concatenate multiple methods, if need.

pixtur (changes requested):
Please check comment regarding multiinput

PR #767, opened 2025-09-05
Audio waveform operator

stereomuviz:
Add operator to read audio waveform from left and right channels


pixtur:
Thanks ago for the PR!

Some feedback that surfaced while reviewing / merging:

- There is a set of methods only related to computing the wave form data. I moved that and all related private data into a new static class `WaveFormProcessing`
- I clarified the names of some parameter and method names.
- Please avoid defining multiple variables: e.g. `float a,b;`
- The computation of the waveform frequencies is not trivial. I changed the implementation so that it's only done, if required by the [WaveForm] operator.

PR #768, opened 2025-09-07
Improve parameters window layout

newemka:
1. Remove redundant namespace
2. Works better with Ui scaling
3. Enabled and Bypass button have consistent size 
4. Enabled and Bypass button are alway visible 
TiXL_qsemAETbkI


pixtur:
Nice!

PR #773, opened 2025-09-12
Add Volume Waveform to Timeline 

newemka:
It's still a proof of concept. 

I've added a button to TimelineControls.cs, it's using Icon.Stripe4PxPattern
image
Click this button to switch between Spectrum, Waveform or Both.

I have the feeling that generating the image is too slow compared to generating only the original spectrum. 
Maybe some optimisation can be done. 

I've tried to find colors that will look okay in any of the default color themes. 

TiXL_nMfzUtvI9l


sense-amr:
I think this colour selection is the most pleasing

newemka:
As asked on discord: "Will it be possible to deactivate those visual representations completely?"
Thus I've added DisplayMode.Duration  
TiXL_a7WzsGiKka

ylmrx:
Regarding the palette, I think the default should be some gradient across some carefully picked theme colors, so they "work fine" in any theme.

Custom hard-coded ones are a very nice addition.

newemka:
This feature has been implemented in a different way in 4.3.  I can close this.

PR #784, opened 2025-09-17
Clear Keyframe selection when draggin playhead

newemka:
Issue: Selection clearing consistency between clicking in the timeline and using the dragging area.
When using the dragging area the keyframe selection is not cleared but duplicate keyframe is not working. 
TiXL_jpTL3OjfyI

this pull request adds selection clearing when using the dragging area: 
TiXL_Fd8tEVlvRG

Of course this is a temporary solution, a better solution would be to handle the selection clearing in an other way.

### DIFF
--- a/Editor/Gui/Windows/TimeLine/TimeLineCanvas.cs
+++ b/Editor/Gui/Windows/TimeLine/TimeLineCanvas.cs
@@ -273,6 +273,12 @@ internal sealed class TimeLineCanvas : CurveEditCanvas
             Playback.TimeInBars = draggedTime;
         }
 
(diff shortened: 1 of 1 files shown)


pixtur:
I agree that something seems to be off with the ctrl+D shortcut to duplicate the keyframe selection. 
But this PR doesn't seem to be an ideal solution because it changes the way keyframe interaction works.

As a work-around you can use the Context-menu -> Duplicate Selected Keyframes for now.

newemka:
You're right, the context menu is working great. Thanks! I can close this PR.

PR #785, opened 2025-09-20
Fix: [DrawMesh] Flat Shading

newemka:
Flat Shading render was all black, this PR brings back the flat shading render.


pixtur:
Good Catch!

PR #789, opened 2025-09-23
Update EmguCV to OpenCVSharp and VideoDeviceInput. Add CameraCalibrator and Video2DPointsScan

banidu-512:


Operators/Lib/Lib.csproj
```diff
@@ -9,13 +9,13 @@
     <AssemblyName>$(RootNamespace)</AssemblyName>
     <Configurations>Debug;Release</Configurations>
     <Platforms>AnyCPU</Platforms>
-    <VersionPrefix>1.666.999</VersionPrefix>
+    <VersionPrefix>1.668.999</VersionPrefix>
     <HomeGuid>00000000-0000-0000-0000-000000000000</HomeGuid>
-    <EditorVersion>$(TixlVersion)</EditorVersion>
+    <EditorVersion>4.1.0</EditorVersion>
```
pixtur, line 14: This is not good and should be reverted.
Please make use to keep the original property unchanged for all csproj files because this makes sure that it will work for upcoming versions.
The csproj are changed automatically on TiXL start, but the changes should not be committed!

Operators/Lib/Lib.csproj
```diff
@@ -101,6 +101,9 @@
     <PackageReference Include="DirectShowLib.Standard" Version="2.1.0" />
     <PackageReference Include="Emgu.CV" Version="4.11.0.5746" />
     <PackageReference Include="Emgu.CV.runtime.windows" Version="4.11.0.5746" />
```
pixtur, line 103: So, if openCvSharp4 replaces Emgu, that this should be removed here, right?
We don't have to have unnecessary dependencies.

Operators/Lib/Lib.csproj
banidu-512: Yes sir! i just missed it.

Operators/Lib/Lib.csproj
banidu-512: Yes! Theres no any other node luckily.

PR #791, opened 2025-09-23
New nodes: VideoStreamInput and PosiStageNet support

banidu-512:


pixtur:
Wow. This looks like a super cool feature set! But I think I need a little bit more context about this PR. What do the ops do?

Also, it would be great to have [more descriptions](https://github.com/tixl3d/tixl/wiki/dev.OperatorConventions#checklist) in the ops and their parameters. I know that this is painful and takes a lot of time, but in my experience, if an op doesn't have a description, it will not be discovered and used by other people than the original author.

Every time I took the time to write documentation, set parameter ranges, and create presets, the overall quality of the operator got better because I found further possibilities to tweak it.

I would also really love to separate out everything that relies on Python. I love Python as a programming language, but as an ecosystem I think it's even worse than .NET. You will need *a lot* of good arguments to convince me that it should be installed by default with the TiXL installer. If it shouldn't, then the lib shouldn't have any dependencies on it.

pixtur (changes requested):
See comment.

banidu-512:
> Wow. This looks like a super cool feature set! But I think I need a little bit more context about this PR. What do the ops do?
> 
> Also, it would be great to have [more descriptions](https://github.com/tixl3d/tixl/wiki/dev.OperatorConventions#checklist) in the ops and their parameters. I know that this is painful and takes a lot of time, but in my experience, if an op doesn't have a description, it will not be discovered and used by other people than the original author.
> 
> Every time I took the time to write documentation, set parameter ranges, and create presets, the overall quality of the operator got better because I found further possibilities to tweak it.
> 
> I would also really love to separate out everything that relies on Python. I love Python as a programming language, but as an ecosystem I think it's even worse than .NET. You will need _a lot_ of good arguments to convince me that it should be installed by default with the TiXL installer. If it shouldn't, then the lib shouldn't have any dependencies on it.

i think i had made mistakes with the git stuffs. it seems like more thing came here that i wanted in the PR. basicallt these two thhat was planned originally here:
@banidu-512
[Add [VideoStreamInput]](https://github.com/tixl3d/tixl/pull/791/commits/00a8c4bb15743855f2b6c2bfbce50c05fdedebb7) 
[00a8c4b](https://github.com/tixl3d/tixl/pull/791/commits/00a8c4bb15743855f2b6c2bfbce50c05fdedebb7)
@banidu-512
[Add [PosiStageInput], [PosiStageOutput]](https://github.com/tixl3d/tixl/pull/791/commits/38b5e176cadf7eb147034c4ba244ed093bf43424) 
[38b5e17](https://github.com/tixl3d/tixl/pull/791/commits/38b5e176cadf7eb147034c4ba244ed093bf43424)

Im going to sit down on the nodes to fill it up with documentation.
for the python thing i can start with this example: https://forum.vvvv.org/t/vl-pythonnet-and-ai-worflows-like-streamdiffusion-in-vvvv-gamma/22596

pixtur:
@banidu-512 sorry. Reviewing this tool longer than expected. Super cool contribution!

Some notes: Please avoid implementing `IDisposible` in ops:


```cs
    internal sealed class PosiStageOutput : Instance<PosiStageOutput>, IStatusProvider, ICustomDropdownHolder, IDisposable ❌
    {
```    

```cs
    internal sealed class PosiStageOutput : Instance<PosiStageOutput>, IStatusProvider, ICustomDropdownHolder ✔
    {
```    

The Instance already implements that internally.

so basically this...

```cs
        public void Dispose() // ❌ incorrect
        {
            StopListening(); 
            _outputBuffer?.Dispose();
        }
```

should be...
```cs
        protected override void Dispose(bool isDisposing)
        {
            if (!isDisposing)
                return;

            StopListening(); 
            _outputBuffer?.Dispose();
        }
```


Read more here: https://github.com/tixl3d/tixl/wiki/dev.DevelopingOperators#implementing-the-correct-dispose-pattern


Also avoid unnecessary usings:

image

Regarding code formatting please consider the coding convention:

although this _is_  actually readable...

```cs
                            if (reader.BaseStream.Position + fieldLen > trackerEndPos) break;

                            if (fieldId == 0x0000) newPos = new Vector3(reader.ReadSingle(), reader.ReadSingle(), reader.ReadSingle());
                            else if (fieldId == 0x0002) newOri = new Vector3(reader.ReadSingle(), reader.ReadSingle(), reader.ReadSingle());
                            else reader.BaseStream.Position += fieldLen;
```                            

if would prefer if(...) {...} or even...

```cs
                            switch (fieldId)
                            {
                                case 0x0000:
                                    newPos = new Vector3(reader.ReadSingle(), reader.ReadSingle(), reader.ReadSingle());
                                    break;
                                case 0x0002:
                                    newOri = new Vector3(reader.ReadSingle(), reader.ReadSingle(), reader.ReadSingle());
                                    break;
                                default:
                                    reader.BaseStream.Position += fieldLen;
                                    break;
                            } 
```

PR #792, opened 2025-09-23
Add [CustomPython] operator.

banidu-512:
-Python implementation with Python.NET

### DIFF
--- a/Core/Core.csproj
+++ b/Core/Core.csproj
@@ -25,6 +25,9 @@
         <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
         <PackageReference Include="NuGet.Configuration" Version="6.13.2" />
         <PackageReference Include="pythonnet" Version="3.0.5" />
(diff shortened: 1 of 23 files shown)


pixtur:
Honestly, I'm not sure if adding a python dependencies is a good thing for TiXL.

I _can_ see a scripting language implementation for something like LUA, because we can fully embed it without additional dependency. But for something like Python, it will add a whole different suite of potential problems. I wouldn't be qualified (or wouldn't have a time) to support people on installation issues because their have an outdated version of PiP or some environment variable mismatch.

I think this would be a great example for a package that can be installed additionally to TiXL from another source (similar to Nuget packages). We could look into VVVV and check out their package system (which is also using nuget).

Let's discuss.

brainrake:
This PR doesn't add Python itself as a dependency, but Python.NET, a wrapper library, and the user has to bring their own snake. But the maintenance burden is real. A case for out of tree modules?

pixtur:
I merged the other changes because I think they are highly useful. For Python itself, I skipped it for now.

PR #793, opened 2025-09-25
Add: [HSBToColor]

newemka:
[HSBToColor] gives same results as HSL(renamed to HSB) from ColorEditPopup.cs AKA color picker
image


newemka:
Preventing from resetting to black when S, B and Value slider are reaching 0
TiXL_rflO0uLPta

pixtur (changes requested):
Super cool fix! Slight tweaks requested.

pixtur:
Awesome!!!

PR #794, opened 2025-09-25
OutputWindow Scrollbar for toolbar

newemka:
The Scrollbar appears if the output window is not wide enough to display the whole toolbar. Allowing to reach the hidden buttons.


pixtur:
nice

PR #799, opened 2025-10-08
Normal Map support for RaymarchField

newemka:
image


newemka:
Comparing the normal map on a mesh with SDF 
image
the tangent needs to be fixed: 
here i'm testing with` tangentNormal.xy = -tangentNormal.xy;`
image

newemka:
I baked a normal map with Blender and using it to make experiments 
TiXL_sLOPWMyxPs
this looks right. But it's still wrong when I look at the other sides of the cube...

pixtur:
Absolutely awesome. I will clean this up a bit after merging.

PR #808, opened 2025-10-17
Add [DataPointImportExport] and [DataPointConverter] for managing pointdata

banidu-512:
- DataPointImportExport:
-  - Imports/exports JSON point lists to/from a GPU structured buffer.  
-  - Features an autoload capability for import path changes.   - Retains imported data in an internal buffer.

- DataPointConverter: 
-   - Converts point data from CSV or JSON files into a GPU buffer. 
-   - Supports custom column mapping for flexible data parsing. 
-   - Allows exporting converted points back to CSV or JSON. 
-   - Handles both Euler angles and Quaternions for rotation.

### DIFF
--- a/Operators/Lib/point/io/DataPointConverter.cs
+++ b/Operators/Lib/point/io/DataPointConverter.cs
@@ -1,0 +1,432 @@
+﻿#nullable enable
+using System.Globalization;
+using System.Text.Json;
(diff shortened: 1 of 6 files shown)


pixtur:
Interesting. This nicely aligns with a spread sheet feature I have planned for the near future...

PR #809, opened 2025-10-17
Update: [SetEnvironment]

newemka:
_Orientation_ is working now. 
I've added [HasValueChanged] for input parameters connected to [_SpecularPrefilter]
Thus, we are not forced anymore to enable _Update Live_  when editing the following parameters: 

1. Exposure
2. Quality Factor
3. Orientation


pixtur:
Wow! So cool!

PR #811, opened 2025-10-18
Change wording from Tooll to TiXL aswell as changed redundant wiki links.

catsizedcoder:
I have went through and modified a few links to the newer wiki links, aswell as changed the wording in StartupValidation to TiXL instead of Tooll


pixtur:
Wow excellent!

PR #812, opened 2025-10-18
Fix 810: Adding nodes downstream of alternate outputs shows a wrong connection

### LINKED ISSUE #810: Adding nodes downstream of alternate outputs shows a wrong connection
**Description:**
When a node has more than one output, clicking that output to add a downstream node results in a connection line originating from the "topmost" output.

**System information:**
 - OS: Windows 11
 - Version: v4.0.6.3
 - Graphics card and resolution: RTX 4070 Ti Super, 1920x1080

**To Reproduce:**
1. Spawn a node with more than one output type
2. Add a node downstream of each output
3. Observe upon clicking on a secondary output's connection that the "add node" connection originates from the "topmost" output

**Expected behavior:**
The "add node" connection should originate from whichever output was clicked/whichever output connects to the selected connection.

**Screenshots:**
Image
Image

**Additional context:**
None

deanchumash:
Closes #810

I added a calculation of OutputLineIndex in `PlaceholderCreation.cs` so that the temporary connection line originates from the correct point when splitting a connection.

A later issue may fix the persisting connection line origin mismatch when making a placeholder for a disabled output type.

w2x6XhlHXe

### DIFF
--- a/Editor/Gui/MagGraph/Interaction/PlaceholderCreation.cs
+++ b/Editor/Gui/MagGraph/Interaction/PlaceholderCreation.cs
@@ -39,10 +39,16 @@ internal sealed class PlaceholderCreation
 
         var outputLineIndex = hoverConnection.SourceItem.Instance?.Outputs.IndexOf(hoverConnection.SourceOutput) ?? 0;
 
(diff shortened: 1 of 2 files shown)


pixtur:
Wow! Looks really good!

PR #813, opened 2025-10-18
Ui: NewSubfolder Window looks a bit better

newemka:
image


pixtur:
I really should start with the new asset file browser...

PR #814, opened 2025-10-18
Adds DrawOperatorDragHandle to AudioReactionUi

newemka:
Before: the audio values visualizer is drawn outside of the operator (bottom left)
image

After: 
image


pixtur:
Good catch! Much cleaner.

PR #815, opened 2025-10-18
Add [FreeDInput],[FreeDOutput] operators for FreeD protocol

banidu-512:
Implements operators for sending and receiving camera tracking data over UDP using the FreeD protocol.

*   `FreeDOutput` sends camera position, rotation, zoom, focus,
    and user-defined data.
*   `FreeDInput` listens for and parses incoming FreeD packets,
    providing outputs for individual camera data (position,
    rotation, zoom, focus, user) and a buffer for multiple
    cameras suitable for GPU processing, along with a dictionary
    of all tracked cameras.

This functionality is essential for virtual production, augmented reality, and broadcasting, enabling real-time integration with external systems like Unreal Engine, Vizrt, or Aximmetry.

### DIFF
--- a/Operators/Lib/io/freed/FreeDInput.cs
+++ b/Operators/Lib/io/freed/FreeDInput.cs
@@ -1,0 +1,357 @@
+#nullable enable
+using System.Net.NetworkInformation;
+using System.Net.Sockets;
+using System.Threading;
+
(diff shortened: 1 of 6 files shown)


pixtur:
Wow! So cool. Thanks for also adding documentation!

PR #826, opened 2025-10-26
ParameterWindow Ctrl+LMB on symbol name to copy to clipboard

newemka:
This is useful for communication, when helping someone on Discord for example.


pixtur:
I think having different options to copy information from the context menu would be more versatile.

PR #851, opened 2025-11-12
Add: Screen Manager

newemka:
image

### DIFF
--- a/Editor/App/AppWindow.cs
+++ b/Editor/App/AppWindow.cs
@@ -16,7 +16,6 @@ using Icon = System.Drawing.Icon;
 using Rectangle = System.Drawing.Rectangle;
 using Resource = SharpDX.Direct3D11.Resource;
 using Vector2 = System.Numerics.Vector2;
-using Vector4 = System.Numerics.Vector4;
 
 namespace T3.Editor.App;
 
--- a/Editor/App/ProgramWindows.cs
+++ b/Editor/App/ProgramWindows.cs
@@ -7,13 +7,13 @@ using SharpDX.DXGI;
 using T3.Core.IO;
 using T3.Core.Resource;
 using T3.Core.SystemUi;
-using T3.Editor.Gui;
+using T3.Editor.Gui; // for ReleaseMode
 using T3.Editor.Gui.UiHelpers;
 using T3.Editor.UiModel;
 using Device = SharpDX.Direct3D11.Device;
 using PixelShader = T3.Core.DataTypes.PixelShader;
 using VertexShader = T3.Core.DataTypes.VertexShader;
-using Vector4 = System.Numerics.Vector4;
+using Vector2 = System.Numerics.Vector2;
 
 namespace T3.Editor.App;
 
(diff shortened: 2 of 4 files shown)


Editor/App/AppWindow.cs
```diff
@@ -16,6 +16,7 @@
 using Rectangle = System.Drawing.Rectangle;
 using Resource = SharpDX.Direct3D11.Resource;
 using Vector2 = System.Numerics.Vector2;
+using Vector4 = System.Numerics.Vector4;
```
pixtur: Try to use ImRect instead of Vector4.

Editor/Gui/Windows/ScreenManager.cs
```diff
@@ -0,0 +1,441 @@
...
+using T3.Editor.Gui.Input;
+using T3.Editor.Gui.Styling;
+using T3.Editor.Gui.UiHelpers;
+using T3.Editor.Gui.Windows.Layouts;
+using Icon = T3.Editor.Gui.Styling.Icon;
+using Vector2 = System.Numerics.Vector2;
+
+namespace T3.Editor.Gui.Windows;
+
+internal sealed class ScreenManager : Window
```
pixtur:
```suggestion
internal sealed class ScreenManagerWindow: Window
```

Editor/Gui/Windows/ScreenManager.cs
```diff
@@ -0,0 +1,441 @@
...
+{
+    internal ScreenManager()
+    {
+        Config.Title = "Screen Manager";
+        SystemEvents.DisplaySettingsChanged += (_, __) => _layoutDirty = true;
```
pixtur: Careful: If you subscribe to an event like, this you normally would need to implement `IDispose` to clean this up. Here it's not really a problem, because there is only one Window. But event handlers can be tricky and ask to be used with care.

Editor/Gui/Windows/ScreenManager.cs
```diff
@@ -0,0 +1,441 @@
...
+        var spanningBounds = UserSettings.Config.OutputArea;
+
+        // Check if spanning area is defined
+        if (spanningBounds.Z > 0 && spanningBounds.W > 0)
+        {
+            // Enable the output window if not already enabled
+            if (!WindowManager.ShowSecondaryRenderWindow)
+            {
+                WindowManager.ShowSecondaryRenderWindow = true;
+            }
```
pixtur: Why check if you can just set?
```suggestion
WindowManager.ShowSecondaryRenderWindow = true;
```

Editor/Gui/Windows/ScreenManager.cs
```diff
@@ -0,0 +1,441 @@
...
+        _screensInSpanning.Clear();
+        foreach (var s in _cachedScreens)
+            if (IsScreenInSpanningArea(s, _cachedSpanning))
+                _screensInSpanning.Add(s);
+    }
+
+
+    // -------------------------------------------
+    // Helper: Compute scale to fit screens in UI
+    // -------------------------------------------
```
pixtur:
```suggestion
    /** Helper: Compute scale to fit screens in UI  **/
```

Please use consistent comment style. Using the .net style will make documentation available on hover.

Editor/Gui/Windows/ScreenManager.cs
```diff
@@ -0,0 +1,441 @@
...
+
+        var finalScale = baseScale * MathF.Min(1, scaleFactorX);
+        finalScale = Math.Clamp(finalScale, .6f * baseScale, 2.0f);
+
+        return finalScale;
+    }
+
+    // -------------------------------------------
+    // DrawScreenLayout() - optimized for low CPU
+    // -------------------------------------------
```
pixtur: see above.

pixtur (changes requested):
Excellent PR! Some comments below (mostly regarding style and GC)

pixtur:
Awesome!

PR #852, opened 2025-11-12
Fix: Snapping velocity, SmoothItemPosition() and brings keyboard navigation to MagGraph (perfect for laptops without mouse)

newemka:
Snap is really "snappy", nodes position damping remains smooth.  
```
const float normalDampAmount = 0.6f;
const float snapDampAmount = 0.2f; // Much snappier
```

About WASD MagGraph navigation: it's here because I forgot to remove it before pushing to my fork

### DIFF
--- a/Editor/Gui/MagGraph/Ui/MagGraphCanvas.Drawing.cs
+++ b/Editor/Gui/MagGraph/Ui/MagGraphCanvas.Drawing.cs
@@ -2,6 +2,8 @@ using ImGuiNET;
 using T3.Core.DataTypes.Vector;
 using T3.Core.Operator.Interfaces;
 using T3.Core.Utils;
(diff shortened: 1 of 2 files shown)


pixtur:
To be honest, I think that WASD will always feel awkward for graph scrolling and it blocks a lot of potentially important keyboard-shortcuts. This feels like a potential [XY-Problem](https://en.wikipedia.org/wiki/XY_problem), I.e. windows trackpad feels horrible, so let's build a work around. The problem is that horrible trackpad and should be fixed.

I will apply your damp setting separately.

newemka:
@pixtur Yes, sorry I made a mess with this WASD inside this PR.

pixtur:
No worries.

PR #854, opened 2025-11-13
Add history window which allows to see the state of the undo stack

mrvux:
As the title says, this is just a simple window that shows the state of the undo stack.

This is really useful to know what were last commands, also for debug to double check that some commands go in stack or not

### DIFF
--- a/Editor/Gui/Windows/History/HistoryWindow.cs
+++ b/Editor/Gui/Windows/History/HistoryWindow.cs
@@ -1,0 +1,46 @@
+using ImGuiNET;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using T3.Editor.UiModel.Commands;
+
+namespace T3.Editor.Gui.Windows.History;
+
+internal sealed class HistoryWindow : Window
+{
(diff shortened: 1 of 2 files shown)


pixtur:
I know it's not obvious and too hidden, but this code (literally the same) already exists in  `UtilitiesWindow.cs`

That would be TiXL->Debug->Utilities in the app menu.

```cs
                case Categories.DebugInformation:
                    FormInputs.AddSectionHeader("Debug information");

                    if (ImGui.TreeNode("Undo history"))
                    {
                        var index = 0;
                        foreach (var c in UndoRedoStack.UndoStack)
                        {
                            ImGui.PushStyleColor(ImGuiCol.Text, new Vector4(0.5f / (index + 1) + 0.5f));
                            ImGui.PushFont(index == 0 ? Fonts.FontBold : Fonts.FontNormal);
                            if (c is MacroCommand macroCommand)
                            {
                                ImGui.Selectable($"{c.Name} ({macroCommand.Count})");
                            }
                            else
                            {
                                ImGui.Selectable(c.Name);
                            }

                            ImGui.PopFont();
                            ImGui.PopStyleColor();
                            index++;
                        }

                        ImGui.TreePop();
                    }
```

PR #855, opened 2025-11-13
Update: MagGraph CreatePlaceHolderConnectedToInput

newemka:
Problem:
If the selected operator is too close to the left edge of the Graph view, adding a new operator by clicking on the input of a parameter will add the place holder outside of the graph view. 
Solution:
Make sure the place holder is visible by scrolling the view to the right. 
TiXL_2aFvaVVn06


pixtur:
Awesome micro-optimization. I actually fell into that trap a bunch of times!

PR #857, opened 2025-11-16
Add: FXAA (Fast approXimate Anti-Aliasing) 

newemka:
It works well to improve RayMarching where MSAA has no effect. 
TiXL_8JBmpruqBa


pixtur:
That's so awesome!

PR #858, opened 2025-11-17
Fix: MagGraph CreatePlaceHolderConnectedToInput

newemka:
Simplified and works as expected.


pixtur:
Nice quality of live improvement.

PR #859, opened 2025-11-17
Implement project directory list rather than single project directory

deanchumash:
This is step 1 of 2 to add automatic USB project discovery.

Implementation of multiple project directories.

Synopsis of changes:

1. I created a new config item that is a list of strings to hold each directory searched for projects.
2. I added an EditableListBox FormInput (https://github.com/deanchumash/tixl/blob/list-usb-projects/Editor/Gui/Input/FormInputs.cs#L315)
3. I updated each reference to the config string for project directory to point to the newly created list.
!!! I left a few TODOs in places where I thought it would be important to change which item of the project directory is being used !!!
4. I added some logic in program startup to populate this newly created list with the default directory (Documents/TiXL) if it is empty.

Please test further before merge, as I only tested this on two of my own devices which previously had TiXL installed.. not sure what will happen on a fresh computer, but I don't have one as of right now. But as it stands, works nicely and saves between restarts.

### DIFF
--- a/Editor/Gui/Input/FormInputs.cs
+++ b/Editor/Gui/Input/FormInputs.cs
@@ -450,7 +450,7 @@ internal static class FormInputs
                  // Compute reserved right width same as per-item (X + + spacing) so plus aligns exactly
                  var styleBottom = ImGui.GetStyle();
                  var padXBottom = styleBottom.FramePadding.X * 2;
-                 var wXBottom = ImGui.CalcTextSize("X").X + padXBottom;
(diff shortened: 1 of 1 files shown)


deanchumash:
#860 is dependent on this being merged.

pixtur:
Oh dear. I think I've merged the PRs in the wrong and now have a merge conflict.
Maybe we can have quick chat on Discord to clean this up.

deanchumash:
All good. #860 was branched off of this, so it included these commits.

PR #860, opened 2025-11-18
Automatic detection of USB drives with config option

deanchumash:
This is step 2 of 2 to add automatic USB project discovery.
This merge request depends on #859

Implementation of multiple project directories.

Synopsis of changes:

1. I created a new config item for whether USB detection is enabled
2. I added logic in program startup to detect projects in the `TiXLProjects` folder of any removable media
3. I added a new USB glyph to show in the Project Hub if a project is on removable media. It may be helpful to add this to the glyph to the Symbol Library.

image


pixtur:
Wow. That looks really great! 
I will try to test this setup carefully.

PR #864, opened 2025-11-21
Add: Copy Selected Symbol Name to Clipboard to MagGraph context menu

newemka:
Useful when editing descriptions or telling someone about a specific symbol.
image


pixtur:
I would be great combine the different copies into a "Copy >" Menu group.
I frequently would love to copy these things:
- `SymbolId`
- `SymbolChildId`.
- List of other Symbols using it
- List of other Symbols being used

We could even have options to...
- Search for that filter on Symbol library
- Filter Console log messages of that instance (Although I'm not sure if that's really a use-case)

PR #865, opened 2025-11-22
Update / clean / unify: PickFromList operators

newemka:
1. `PickFloatFromList` was missing the modulo. 
2. Renaming  `PickFromStringList` to `PickStringFromList` for consistency
3. Unifying code across the three operators


pixtur:
Oh nice! So much better!

PR #878, opened 2025-12-03
Fix: HasValueDecreased

H445:
Fixed HasValueDecreased to properly show output value on subsequent decreases in value


pixtur:
Good catch!

PR #883, opened 2025-12-06
fix: crash when returning to audio clip timeline without a defined audio input

ylmrx:
Reproducer:

- Enable project audio
- In `Project Sound Track` Select some audio clip
- Switch to `External Device`, pick some external audio interface
- Save & Exit
- Unplug audio interface
- Go back to `Playback Settings` (Audio interface picker is empty)
- Return to project sound track, Tixl crashes

This check != null fixes it.


pixtur:
Good catch! Thanks for fixing.

PR #884, opened 2025-12-07
feat: Add ZipStringList, and LaunchPadDisplay

ylmrx:
- ZipStringList : new operator allowing to zip two lists of strings, such as:

```txt
zip([a,b,c,d], [1,2,3]) => [a,1,b,2,c,3]
```

- LaunchPadXDisplay : convert a texture2D to a sysex message to display it on a LaunchPad X RGB pads (It can be adapted to more similar devices)

- RGBToSysex : Convert a flattened ternary (ie. `[R1,G1,B1,R2,G2,B2,...,Rn,Gn,Bn]`) list of RGB [0-255] floats to a space-separated string their rounded integer parts converted to hexadecimal, such as:

```txt
[255.0, 0.0, 0.2, 0.9, 254.8, 0.0] => "ff 00 00 01 ff 00"
```

This is to be used along MidiSysexOutput.

### DIFF
--- a/Operators/Lib/string/list/ZipStringList.cs
+++ b/Operators/Lib/string/list/ZipStringList.cs
@@ -16,26 +16,16 @@ internal sealed class ZipStringList : Instance<ZipStringList>
     private void Update(EvaluationContext context)
     {
         var strOne = StringsOne.GetValue(context);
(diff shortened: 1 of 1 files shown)


H445:
This could be simplified using System.Linq.Zip().

PR #885, opened 2025-12-10
Update/Fix: Search results, relevancy weighting, pascal case match

H445:
- Removed the hard cap of 30 search results (Default limit 0 is now unlimited) 
  - This still allows for setting a defined limit for classes like RandomPromptGenerator.cs
- Updated relevancy weighting with less variation between string match types
- Fixed pascal case match to take into account the last letter typed in search
- Filtered search results to only show relevant results (score of 10+)
  - This still allows searching for obsolete and "_" ops directly with string matching, but filters out of generic search
- Moved relevancy boost by usage amount to the bottom, only boost if the result is already relevant (score of 10+)
  - This avoids boosting operators that are used a lot (like Text) above the most relevant results


pixtur:
Excellent update. I'm really curious to test this.

pixtur:
I'm not really sure, why this was flagged as draft.

Issue #886, opened 2025-12-12
Crash: Bug in latest RTV update v4.1.0.8-alpha a2e503


H445:
**Describe the bug**
Excessive error messages when trying to evaluate a node before a render node
Crashing many projects

**Please complete the following information):**
 - OS: Win11
 - Version: v4.1.0.8-alpha a2e503
 - Graphics card and resolution : N/A

**To Reproduce**
1. Evaluate any node before a render target
<img width="1308" height="779" alt="Image" src="https://github.com/user-attachments/assets/1560a220-f0bf-43ed-a41a-da4f9dd078fa" />
2. See errors.

3. This is also causing crashes trying to open nodes in saved project files / sometimes simply when opening the project

[message (4).txt](https://github.com/user-attachments/files/24136592/message.4.txt)

<img width="1162" height="652" alt="Image" src="https://github.com/user-attachments/assets/260c8f1b-89b8-45f6-ac48-14adff6ea33e" />

**Expected behavior**
No errors or crashing

**Screenshots**
See above

H445:
fix for _SpecularPrefilter (This fixes the crash, but the errors are still there.)
```
        foreach (var rtv in _prevRenderTargetViews)
        {
            rtv?.Dispose();            
        }
```

H445:
PR for crash fix: https://github.com/tixl3d/tixl/pull/887

PR #887, opened 2025-12-12
Fix: rtv dispose crash and excessive logging #886

H445:
All fixes for: https://github.com/tixl3d/tixl/issues/886


H445:
Everything fixed from https://github.com/tixl3d/tixl/issues/886

pixtur:
Excellent! Thanks for fixing.

One thing I noticed was, that we could also check to rtv.IsDisposed before disposing.

PR #888, opened 2025-12-13
fix: fixes #873, do not assert OutputLines size on Output operator

### LINKED ISSUE #873: Crash: Dragging from Output node connection
**Describe the bug**
Tixl will crash if you try to click and drag from the connection node of an Output operator in a subgraph

**Please complete the following information):**
 - OS: Windows
 - Version: 4.1.0.8a

**To Reproduce**
1. Create a subgraph with an Output node
2. Attempt to drag a connection (backwards) from the output node

**Expected behavior**
This should allow you to select a compatible operator out of a list to attach, having the same functionality as other node types.

ylmrx:


pixtur:
Excellent. Good catch!

PR #889, opened 2025-12-14
Fix: CustomPixelShader Graph - GetTextureSize

H445:
Fixed No connection to SetResolutionSize:

image

Scaled fallback image to RequestedResolution so it will not be 8x8 in case the resolution is not set and image is not provided

image


pixtur:
Perfect!

PR #890, opened 2025-12-15
Update: ImGuiListClipper implementation for search

H445:
Search now using List Clipper.

Added new helper class to override width of ExtendToLastItem - This stops the clipper from updating to the last items width while scrolling


pixtur:
Thanks looks great!

PR #892, opened 2025-12-16
add LaunchPadXButton, LaunchPadXFader4 and LaunchPadXFader8 + examples

ylmrx:
Iterating upon the SysexOutput operator in a Launchpad controller use-case.
This implements a stateful launchpad X grid-controller integration.

Bring in a new Operator LaunchPadXButton in order to abstract boolean of a Launchpad grid button and its status (RGB on or dimmed).

- reference a button with its `X`/`Y` coordinates
- reflect its state with its RGB LED (`Color` picker, dimmed out when Off)
- `FromState` toggle take status from the `State` boolean input, instead of the MIDI input (cf upper-left metronome display in the example)
- `Init` takes a trigger to reset the grid and switch LED according to State inputs
- `Toggle` switch the behavior between a toggle (flip when push) or an instant (true when pushed) button. LED is updated accordingly

Bring in `LaunchPadXFader4` and `LaunchpadXFader8` operators. Horizontal fader of 4 or 8 pad wide faders, ouput 1-4 or 8 integers, or a normalized-to-one float.

The example shows many use cases.


pixtur:
That sounds awesome. 
I wonder if it could be abstract to work for other MidiControllers too.

I'm reluctantly merging but I have the gut feeling that there are just too many controllers to add examples or special use-cases for all of them. I think it would be cooler to actually extend the MidiAdapter implementations for features like FaderGroups and toggle grids.

PR #899, opened 2025-12-21
fix: add a peak LED on signal display, avoid overflow

ylmrx:
- ensure bass api level is >0 (signed integer overflowing on very loud music and full-scale white noise)
- replace slider with a custom meter:
  - use better padding (meter went beyond panel limits on overdriven signals)
  - add a peak LED
  - add some ballistic to smooth out level decay
- fix few labels


ylmrx:
https://github.com/user-attachments/assets/8e0431a8-8998-48f8-98ac-418bca4462e2

(music snippet , Jomaa - Free Bird (Amelie Lens 016 mix-tape))

pixtur:
Awesome! Looks so much better

PR #901, opened 2025-12-29
Float/DMX Updates and Network Resilience

banidu-512:
New MergeFloatLists Op: Added support for Append, HTP, LTP, FailOver, and Average modes with MaxSize control.

DMX Enhancements: Refactored PointsToDMXLights with 16-bit color support, Pan/Tilt offsets, and visualization axis mapping.

Network Stability: Added 2s retry logic to sACN/ArtNet and implemented dynamic network interface refreshing for ArtNet outputs.

### DIFF
--- a/Operators/Lib/io/onvif/OnvifCamera.cs
+++ b/Operators/Lib/io/onvif/OnvifCamera.cs
@@ -1,0 +1,1077 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Net.NetworkInformation;
+using System.Net.Sockets;
+using System.Numerics;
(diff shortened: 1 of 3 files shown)


pixtur:
I've never heard about onvif for now. But I assume this can be super useful.

pixtur:
I'm not sure, if you're are aware of this, but changing the GUID of PointsToDMXLights will break all references and connections to this op. 
It's basically identical to removing that Symbol and creating a new completely different one with the same name.

Considering that many of the input Ids did not change, this might be by design. But it's very unfortunate because it's a breaking change that will basically render all project that used that op invalid.
The worst part is, that the operator will just "vanish" including all its connectsion. That will make it very hard to repair or replace it with the new version.

I'm not really sure how to handle this situation. If revert the GUID on main, but this will break all your projects...

PR #902, opened 2025-12-29
Fixed leaving focus mode (F12)

0xNULLderef:
Restores previous functionality of the focus mode without impacting new Skill Quest stuff.
`LoadAndApplyLayoutOrFocusMode()` setting `UserSettings.Config.FocusMode` is a questionable choice and I suggest looking over it, but this works for now.


pixtur:
Nice! I think the whole approach to saving the UI-Layout in user settings should eventually be rethought. 
But is already a great fix.

PR #905, opened 2025-12-31
fix: GetLevel issue the meter with both channel in a single int, only right chan is measured

ylmrx:
## problems

- input level monitored only the right channel
- extremely loud and phasy sounds measured negative value (should be only positive numbers) (good reproducers here are : Noisia — Tommy's Theme drops (phase), Skrillex — Bangarang EP (extreme RMS/brickwalling), full-scale white noise, gain=1)

## fixes

- address `GetLevel()=-1` error report, we should not use those value
- calculate mid+abs(side) signal from both channel (mean) according to upstream doc
- scale it like before


pixtur:
Good catch! I was always irritated by that negative level output but could never narrow down the problem.

PR #908, opened 2026-01-05
Update: adding SpecularAA to [DrawMesh]  

newemka:
Compute normal variance using screen-space derivatives and increase roughness accordingly.
This reduces specular aliasing on silhouettes and high-frequency normalmap regions.
TiXL_d3zrgfARLu

When I made the first proof of concept I was updating **pbr-render.hlsl** and **mesh-Draw.hlsl** , this was breaking every Draw operators who are using **pbr-render.hlsl** until I update them.
With this version I only update **mesh-Draw.hlsl** 
While writing this comment, I'm thinking that the first approach was maybe better.


pixtur:
Super nice. I will try to apply this to other PBR shaders.

PR #913, opened 2026-01-10
fix cs multidispatch call index

unc-quite:
sending the right index value for i==0 (does not affect most common case of callCount==1)


pixtur:
Thanks!

PR #914, opened 2026-01-10
feat: [EaseKeys], [EaseVec2Keys], [EaseVec3Keys] operators

enderprism:
these operators read the keyframes of their <kbd>Value</kbd> field to interpolate values, essentially overwriting the easing you use in the keyframe curves.
it's more ergonomic than the [Ease] family of operators if you're doing keyframed animation.


pixtur:
Nice contribution!

PR #916, opened 2026-01-11
Delete symbol update restrictions

H445:
Running in debug mode now allows deletion of all symbols without restriction
Release mode enforces restrictions as expected


pixtur:
Nice one!

PR #919, opened 2026-01-11
Annotations - allow copy/paste

H445:
Allows copy/pasting to and from annotations

KeyboardActions: 
- Prevent node graph copy/paste if a text input is active (e.g., annotation description)

AnnotationRenaming:
- Added copy/paste ability
- Uses exponential text buffer, allows for pasting large amounts of text (Like node graph json) while keeping overhead as low as possible. Minimum buffer is 256 chars as before.


pixtur:
Nice little detail fix!

PR #922, opened 2026-01-11
[EaseKeys] enhancements

enderprism:
- Fixes the output value of [EaseKeys] operators when the playhead is outside the keyframe range, especially when seeking outside that range (i forgot an early return there, so it still tried to interpolate the wrong value with wrong data).
- Makes it so Constant easing keyframes override the [EaseKeys] easing


enderprism:
the issue seems to still be present with the Vec3 variant, maybe with the Vec2 variant too

enderprism:
ohh right they use loops so they should use `continue` instead of `return`

PR #926, opened 2026-01-12
Fix error message processing in the FileRenderer

ylmrx:


pixtur:
@ylmrx way only a draft?

ylmrx:
It's not finished yet 😄

PR #932, opened 2026-01-14
Add: SNiXL easter egg

H445:
https://github.com/user-attachments/assets/033bfeaa-6222-461b-8ee8-e115fcbe5878


H445:
this is ready to roll unless there are any objections.

pixtur:
Super cute. But I somehow think it would be nice, if the easter egg would not be visible in the AppMenu -> Windows List.

I'm not sure, if this is the case right now.

PR #934, opened 2026-01-15
Update README badges

akx:
Noticed a couple badges still referred to old names (and places).


akx:
> missing the .txt?

Excellent catch 🤦 Fixed (along with the branch name `main`, and removal of the mention of a docusign logo that wouldn't be rendered anyway..)

PR #935, opened 2026-01-18
Refactor path resolution and remove warning

H445:
The AssetFolder constructor was logging a warning whenever a folder path couldn't be resolved.

```Log.Warning($"Can't resolve folder path ? {AliasPath}");```

Problem

This is expected behavior for:
•	The root folder (__root__) which has an empty alias path
•	Virtual folder structures that don't map to physical directories
This caused log spam during normal operation.

Solution
•	Only attempt path resolution when AliasPath is not empty - everything else is already handled..


pixtur:
Good catch and good timing, because I'm in the process of refactoring assets handling.

PR #938, opened 2026-01-20
Fix: [SamplePointColorAttributes] & [AttributesFromImageChannels]

newemka:
Texture Scaling and Gizmo: See [SamplePointColorAttributesExample] before merge to understand the issue
When  [SamplePointColorAttributes] & [AttributesFromImageChannels] are daisy chained, the Gizmos and the scale are not matching, in my fix I had to remove [ApplyTransformMatrix] from both symbols.
Maybe there is something weird with [ApplyTransformMatrix]. 
However, the example works the same after the fix.

Before: 
TiXL_z1K6ZkDXRe

After: 
image


pixtur:
Yes. This was really annoying. Thanks for fixing.

PR #940, opened 2026-01-24
fix: crash upon output creation with timeClip enabled

ylmrx:
## reproducer

- in some custom operator, right-click, add output, fill in, enable "TimeClip"
- click "Add"
- tixl crash
- (if you reopen your project, the output is here and valid)


pixtur:
Good catch!

PR #942, opened 2026-01-26
Update: [VisualizeUvMap] & [DrawMeshUnlit]

newemka:
Add FillMode parameter to [DrawMeshUnlit]

[VisualizeUvMap] uses [DrawMeshUnlit] with wireframe fill mode, this improves a lot its rendering speed for complex meshes.


newemka:
Update: [RepeatMeshAtPoints] add TexCoordFactor parameter.
the goal is to shift TexCoord2 vertically for each copy of the mesh then it's allowing to get unsynchronized VAT animation
TiXL_Maa56bYyac

Big thanks to @unc-quite for the help

PR #946, opened 2026-01-29
AssetLibrary & SymbolLibrary: styling

newemka:
image
Symbol Library:
- uses `UiColors.Text.Fade(0.8f)`

Asset Library: 
- drawing a background when hovering an assest
- the tooltip looks a bit better
- I've referenced the FileVector icon for .svg file 
image
- . svg drag and drop use [LoadSvgAsTexture2D] by default


pixtur:
Nice!

PR #947, opened 2026-01-30
Fix handling of first frame rendering at correct resolution

H445:
There were some cases edge cases, I noticed while testing export a bunch with the audioengine, where it would still skip the first frame on rendering still. There are probably other cases it can happen.

Would still log:
Debug: Video resolution adjusted for codec compatibility: 978x565 -> 978x564
Warning: Skipping frame: waiting for resized frame. Original: 978x565, expected: 978x564, got: 978x565

Added a WaitingForFirstFrame to properly handle the resolution change. It will now wait until it is changed, and set playback time to 0 on frame 0 properly. Also added a counter for debugging failures/timeout


H445:
other issues i found are completely unrelated

pixtur:
Wow. Thanks for looking into this at such detail.

PR #950, opened 2026-02-01
Update: [GainAndBiasExample]

newemka:
+[PlotValueCurve]: I added a bool parameter to enable or disable the label

image


pixtur:
Nice!

PR #962, opened 2026-02-18
feat: [AbsoluteSDF] operator

enderprism:
adds an operator that returns the absolute value of a distance field


pixtur:
Nice one!

PR #966, opened 2026-02-21
Update: [PointsToDMXLights] refactor custom variables to dynamic list…

banidu-512:
… and add per-fixture pan/tilt tracking

- Replace individual SetCustomVar1-16 slots with single CustomVariableChannels list for flexible channel assignment
- Store pan/tilt state per fixture instead of globally to enable proper multi-fixture shortest-path interpolation
- Add HowToDmx example for ArtNet/DMX usage
- Added body for spotlights in VisualizeSpotLights


pixtur:
Thanks for contributing.

PR #969, opened 2026-02-26
Fix: [LoadObj] wrong index order

newemka:
Left: Blender vertex indices 
Right: [LoadObj] vertex indices (SortVertices: Ignore)
image


newemka:
with the last commit, the vertex color is correct even with flat shaded faces image

PR #975, opened 2026-03-08
Update: Parameter's tooltip appears even if animated or has input connected

newemka:
TiXL_B7hG6TD0W3

### DIFF
--- a/Editor/UiModel/InputsAndTypes/InputValueUi.cs
+++ b/Editor/UiModel/InputsAndTypes/InputValueUi.cs
@@ -230,13 +230,13 @@ public abstract class InputValueUi<T> : IInputUi
                 ImGui.PushStyleColor(ImGuiCol.Text, UiColors.ForegroundFull.Rgba);
                 ImGui.Button($"{input.Name.AddSpacesForImGuiOutput()}##ParamName", new Vector2(ParameterNameWidth, 0.0f));
                 ImGui.PopStyleColor();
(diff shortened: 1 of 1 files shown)


Editor/UiModel/InputsAndTypes/InputValueUi.cs
```diff
@@ -297,6 +297,15 @@ InputEditStateFlags DrawConnectedParameter()
...
                     connectedName = "???";
                 }
+                if (ImGui.IsItemHovered())
+                {
+                    var text = "";
+                    if (!string.IsNullOrEmpty(Description))
+                    {
+                        text += Description;
+                    }
+                    CustomComponents.TooltipForLastItem(text);
```
pixtur: I'm confused? Why no....


```suggestion
 if(!string.IsNullOrEmpty(Description) {
          CustomComponents.TooltipForLastItem(Description);  
  }
```

Am I missing something?

newemka: I remember there was a good reason, I need to check :)

newemka: It's because this is the case where there is something connected to the input then we only display the description in the tooltip and no additional note like "Click to reset to default".
I have to admit that this can be improved.

newemka:
with the last commit, when the input is connected, the tooltip display the name of the connected operator or its instance name. 
TiXL_d3hX4HUpCm

PR #980, opened 2026-03-11
Fix: Audio operator evaluation/state management

H445:
The audio player operators were triggering a playback during initial evaluation even if PlayAudio=false. 

example: being evaluated using a switch operator, when switching to the node path including the audio player, it would automatically play back.

Now uses: UpdateStatus and UpdatePlayback instead of a direct Update().

Also tightened up: EnsureOperatorId


pixtur:
Good catch!

PR #981, opened 2026-03-11
Fix: update to align with Roslyn compiler changes in VS2026

H445:
Goal: Proper compilation using VS2026

- Refactor list reversal to avoid LINQ Reverse()
- Updated list reversal logic in ConnectionMaker.cs and Combine.cs to use Array.Reverse and manual for-loops instead of LINQ's .Reverse(). 
- No more dealing with void return type of Reverse, compiles with old/new roslyn versions. 

No functional changes made.


pixtur:
Wow. Thanks for looking into that.

PR #984, opened 2026-03-20
Fix: [Point] initialization with AddSeparator enabled

newemka:
### Problem
When the Point operator had the AddSeparator input enabled in the graph, the buffer was being initialized with the wrong size (1 element) on start. The constructor called UpdateBuffer() before the AddSeparator value could be evaluated, resulting in a buffer sized for a single point instead of the required 2-element list (point + separator).

This caused the CombineBuffers operator to receive incorrectly sized buffers from Points with separators, leading to missing or corrupted data.
As shown here: 

https://github.com/user-attachments/assets/ef7764fe-a4b5-4de8-9b51-7f8c2175e8ec


pixtur:
Oh nice!

Issue #199, opened 2022-10-14
Adding linear interpolation to the sequenceAnim


sIRwa2:
As briefly discussed on the discord, adding  linear interpolation to the sequenceAnim. or creating an ADSR envelope for values. not sure if this shoul dbe 2 seperate requsts :)

![](https://cdn.discordapp.com/attachments/986939795819339777/1030156199146049556/unknown.png)

pixtur:
I think this is actually a very good idea and should be easy to implement.

PR #1026, opened 2026-04-08
Update: [SequenceAnim] interpolation mode

newemka:
In response to "SequenceAnim should support Linear interpolation" #199 
Linear interpolation is fine but I decided to use the easing functions as well. 

image


pixtur:
Oh cool!

PR #1029, opened 2026-04-08
Update: Add support for more/virtual function keys

pixeldesu:
While most keyboards don't have the additional function number row anymore, operating systems at least support them virtually and interpret key presses from them just fine.

While of course more programmatic interfacing is recommended and encouraged with TiXL, if other programs have limited integration support (e.g. no API, just triggering hotkeys), having as many "non-interfering" keys is a plus.

----

Based on this comment in the source
> `/// This enumeration is directly derived from <see cref="System.Windows.Forms.Keys"/>.` 

I went ahead and checked against [the documentation of `System.Windows.Forms.Keys`](https://learn.microsoft.com/en-us/dotnet/api/system.windows.forms.keys) and pulled the appropriate key codes from there.

**Note:** For what it's worth, Dear ImGui / `ImGuiKey` also has support for F13-F24 across implementations, so there shouldn't be any issues here.


pixtur:
Oh darn. That's unfortunate timing. Yesterday, I went ahead and upgraded the ImGui version. This significantly changes the keyboard integration. Currently that's only in https://github.com/tixl3d/tixl/tree/feat/imgui-v1-90 but I'm planning to merge that back to main in the upcoming days.

It would be great, if you could then rebase your PR.

pixtur:
I added the changes manually. 
I hope it will work in the next released version.

PR #1039, opened 2026-04-10
Minimal backup setting

newemka:
I rarely need to use the backup. I still think that it can be interesting to save the project files.

Problem: With full backup my computer is struggling.
Windows defender is scanning the backup zip file and makes my computer barely usable during this time.
This is happening even after adding Tixl's backup folder to the ignore list. 

Solution: I'm using this minimal backup since more than a year. My own version of Autobackup.cs is hardcoded.
Instead of having a 1Go zip file, I get a 3Mo file. 
In this pull request I made it as a user setting that appears only of backups are enabled. 
 
TiXL_QSYElnpcqQ


pixtur:
Good intermediate fix. In the near future we should move the back-archives to the projects, too.

PR #1065, opened 2026-05-30
Fix: per-frame GPU resource leak in _GetSceneDefinitionPoints

newemka:
Stop reallocating BufferWithViews on every update tick; reuse existing instances so SRV/UAV disposal is handled correctly by ResourceManager. Also cache the chunkIndices array to avoid a redundant heap allocation each frame.


pixtur:
Nice!

PR #1091, opened 2026-06-28
Fix right-mouse-click add operator/annotation placement

melw:
Capture position when right-clicking / opening context menu and use it upon creation of operator and/or annotation. Keyboard shortcuts (TAB + Shift+A) stay as they are, using the current active mouse pointer location.


pixtur:
Nice!

PR #1126, opened 2026-07-22
Added "Disconnect" context menu and keyboard shortcut action

juozasg:
Shortcut and context menu disconnects selected nodes from non-selected nodes - just like shaking nodes using the mouse


pixtur:
Very nice! Thanks a lot.

PR #1127, opened 2026-07-22
Restore legacy right-click to select behavior

juozasg:
Uncommented and adapted existing code: right clicking on a non-select node will make it selected (and will open the context menu for the node)


pixtur:
I honestly can remember, why I disabled that code. 🙈

Issue #1137, opened 2026-08-06
Negative offset soundtrack mutes


oteador8:
Negative offset in the Project Sound Track will result in the playback being muted after 0 on the timeline

<img width="553" height="459" alt="Image" src="https://github.com/user-attachments/assets/d9e70ec3-2129-47ed-8b27-69b64a806d54" />

pixtur:
Thanks for reporting!
I thinks this will be fixed with the upcoming version v4.3.

Please reopen this ticket, if the issue remains valid on Main.

PR #1138, opened 2026-08-07
Audio mutes at negative offset 

oteador8:
Negative offset in the Project Sound Track  resulted in the playback being muted after 0 on the timeline
https://github.com/tixl3d/tixl/issues/1137

This seems to fix it.


newemka:
> Negative offset in the Project Sound Track resulted in the playback being muted after 0 on the timeline #1137
> 
> This seems to fix it.

in 4.3 we can add the drag and drop the audio file directly into the timeline and move it to place it as we want.

PR #1155, opened 2026-09-01
feat: DrawTubes operator — full feature set

rofoto:
## Summary

This PR is a total overhaul of the DrawTubes operator that includes,  procedural tube geometry generation with Hermite smoothing, surface rounding, distance-based scaling, and working spin/twist rotation that matches DrawRibbons. In fact you can now set sides to 1 in DrawTubes and get the same behavior (more or less) as DrawRibbons.

### 1. Subdivide Tube sections 
Tube segments can now be subdivided using Hermite curve interpolation for position, rotation, and width/color.

### NEW UI
- **Smooth** (toggle) — enables/disables interpolation
- **SubSegments** (int, 1–10) — number of interpolated rings per source segment

SMOOTH AND SUBDIVING DEMO:

https://github.com/user-attachments/assets/c64f81a4-449b-44db-b4c6-b50499a5c9b3

### 2. Surface Rounding/smoothing
Modifies interpolated tube radius sections using a sine curve. Positive values create bulge/hose shapes, negative values create bead/pearl effects. Source rings remain exact.

### NEW UI
- **RoundAmount** (float, -1 to +1)

ROUNDING DEMO:

https://github.com/user-attachments/assets/1e571f2b-45b6-4ad4-b6a4-ab5392b669ba

### 3. Distance-Based Width Scaling
The original DrawTubes defaulted to scaling the tube based on distance (at least when the input points are coming from pair points for splines)  which was broken (it did not render the last point ever ) — now it can optionally do the same but using the fog distance. 

This feature is a bit fiddly and IMO a bit redundant but I wanted to try to keep old behavior in tact. You can probably achieve the same result by modifying the input points scale based on any arbitrary distance now.  

### NEW UI
- **DistanceScale** (toggle)
- **ScaleNearDist** (float) — full-size distance
- **ScaleFarDist** (float) — min-scale distance
- **MinScale** (float, 0–1)

DISTANCE SCALING DEMO:

https://github.com/user-attachments/assets/5341cc8d-306b-4e42-aeb0-7b17667f2cb9

### 4. Spin & Twist Surface Rotation now working
Rotates the tube's surface orientation (normals + texture UVs) around its length axis so that textures can be used and animated.

### NEW UI
- **Spin** (float, degrees) — constant rotation along entire tube
- **Twist** (float, degrees) — progressive rotation start to end (360 = one full spiral)

SPIN AND TWIST DEMO:

https://github.com/user-attachments/assets/7187e617-ef5d-4e0d-b83a-13315b93597f

### 5. Full Material Support
Tubes now correctly use materials with correct normals.

### 6. Per-Point Width Control
Tube section widths are now controllable using the input points scale or W channel.

### NEW UI
- **UseWAsWeight** (toggle) — use point.W as multiplier
- **UseScale** (toggle) — use point.Stretch.x as multiplier

## Technical Notes

- All geometry procedural in vertex shader via SV_VertexID — no vertex/index buffers
- Draw count computed dynamically in node graph based on Smooth/SubSegments state
- Tension-ready: tangent calc isolated in `getTangent()` for future spline support
- Cbuffer: 24 floats (96 bytes)

## Files Changed

| File | Change |
|------|--------|
| DrawTubes.hlsl | Full vertex shader rewrite with all features |
| DrawTubes.cs | Input slots for all new parameters |
| DrawTubes.t3 | Node graph: inputs, draw-count chain, cbuffer wiring |
| DrawTubes.t3ui | UI layout for all parameters |


rofoto:
conflicts fixes ( just needed to rebase to current main )

PR #1157, opened 2026-09-05
feat: add per-point Space mode to SoftTransformPoints

rofoto:
Adds a `Space` input to `SoftTransformPoints` that switches the transform between
`Object` and `Point` space, letting each point change around its own center
(position, rotation, scale) instead of a shared volume pivot.

Default behavior remains unchanged by sticking to the OBJECT setting. 

DEMO 

https://github.com/user-attachments/assets/c6c2a13f-ad72-4d4a-a44f-2a995e55a13b

- `.cs`: new `Space` input (enum `Spaces { Object, Point }`, default `Object`)
- `.t3`: input registration + wiring to the shader's `CoordinateSpace` parameter
- `.t3ui`: UI toggle for the new input
- `.hlsl`:
  - Object space (unchanged behavior): transforms around the volume
  - Point space: each point spins its own orientation (`p.Rotation = qMul(origRotation, rotFull)`), is displaced in its own local frame (`p.Position += qRotateVec3(Translate, origRotation) * strength`), and scales individually (`p.Scale *= lerp(1, Scale * ScaleMagnitude, strength)`)

### DIFF
--- a/Operators/Lib/Assets/shaders/points/modify/SoftTransformPoints.hlsl
+++ b/Operators/Lib/Assets/shaders/points/modify/SoftTransformPoints.hlsl
@@ -31,6 +31,7 @@ cbuffer Params : register(b1)
     // float UseWAsWeight;
     int VolumeShape; // 28
     int StrengthFactor;
+    int CoordinateSpace; // 0 = Object, 1 = Point
 }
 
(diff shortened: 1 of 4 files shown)


pixtur:
Thanks a lot!

PR #1158, opened 2026-09-05
core: skip dot-folders when scanning package symbols, UIs and source code

rofoto:
Files in .temp were being referenced and conflicting with files in release builds. (this may only effect people developing or using dev/alpha builds) . It bit me while working on my other PRs , leading to new files not being properly updated on rebuilds.

This may or may not be needed with recent move of .temp. 
 
The per-project build output now lives in `.temp/` inside each package folder.
The recursive `EnumerateFiles` scans used for symbols, UIs and source code picked
up those artifacts as real package content (duplicates / stale copies).

Replaces `SearchOption.AllDirectories` with a recursive walker that skips any
sub-directory whose name starts with `.` — the same convention the engine already
uses (`TixlAssemblyLoadContext`). Covers `.temp`, `.git`, and any future dot-folder
automatically.

Applied at 3 sites:
- `Core/Model/SymbolPackage.cs` — symbol search
- `Editor/UiModel/EditorSymbolPackage.cs` — symbol-UI and source-code search
- `Editor/UiModel/EditableSymbolProject.cs` — operator file find

Notes / possible follow-ups:
- The three identical walkers could be extracted into a shared `FileUtils` helper.
- Chose a `.`-prefix rule over matching `FileLocations.TempSubfolder` by name so
  other dot-folders are covered without upkeep.

### DIFF
--- a/Core/Model/SymbolPackage.cs
+++ b/Core/Model/SymbolPackage.cs
@@ -56,7 +56,22 @@ public abstract partial class SymbolPackage : IResourcePackage
             if (!Directory.Exists(dir))
                 return [];
             
-            return Directory.EnumerateFiles(dir, $"*{SymbolExtension}", SearchOption.AllDirectories);
+            return EnumerateSymbolFilesRecursively(new DirectoryInfo(dir));
+        }
+    }
+
+    private static IEnumerable<string> EnumerateSymbolFilesRecursively(DirectoryInfo directory)
+    {
+        foreach (var file in directory.EnumerateFiles($"*{SymbolExtension}"))
+            yield return file.FullName;
+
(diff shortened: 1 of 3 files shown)


pixtur:
Hi, 

to be honest, I'm not really sure I understand the purpose of this PR.
I know, that there was a problem, but I would have thought that we addressed that with 9ceb874a5dbda64681e6d80823959c33d781e202 (moving all symbols to a dedicated Symbols/ folder).

That folder should not contain any .temp folders.

Are you sure, this is still the case?

rofoto:
I think the issue was probably fixed by the PR you mentioned, I just had it happen to me and thought this might be a potential solution, Closed and if I see the same issue going forward will reopen with more info.

PR #1160, opened 2026-09-07
[DrawTubes] Addressing Maintainer Comments & fixes 

rofoto:
## Summary
Follow-up to the [DrawTubes] operator (#1155): reworked texture mapping so UVs have a constant physical texel size, replaced the unused `TextureMode` with a single `UseWAsU` toggle, and added help/tooltip text for every input. This PR also fixes tube caps not rendering correctly (flipped normals) 

### Constant texel-size UVs via arc-length pre-pass
- New compute-shader pre-pass (`DrawTubesArcLength.hlsl`) computes per-segment lengths + an exclusive prefix sum into a cumulative arc-length buffer (fixed-point mm), bound to the vertex stage at `t6`.
- Tube side now maps **U along true arc length** and **V as integer repeats of the circumference**, so the texture stays a constant size along/around the tube instead of stretching with point spacing.
- New **`TextureScale`** input (world units per repeat) sets the texel size for all UVs.
- Sampler wraps on both U and V; end caps are planar disks placed adjacent to each tube end so their density matches the body.
- Falls back to index-based U when there are <2 points or the buffer is unavailable.

### UseWAsU toggle (replaces TextureMode)
The original lifted texturemodes from drawRibbons and many of the options never worked and would have introduced repeated/redundant functionality. 

- Removed the `TextureMode` enum — it was declared but never read in the shader (no effect).
- New **`UseWAsU`** bool: on = U driven by each point's FX1 (interpolated along the tube; caps use endpoint FX1); off = physical arc-length UVs. Useful for animating/slicing a texture with per-point data.

### Documentation
- Tooltips/help added for every input in `DrawTubes.t3ui`; `.md` synced to match the "Export as WIKI" generator output.

### Earlier feedback fixes on this branch
- Operator review feedback (shader/graph/UI/docs) + follow-up UV/cap tweaks.

### Shader warning cleanup
- Removed fxc `X3206` (implicit vector truncation) and `X3203` (signed/unsigned mismatch) via explicit component selection + a `(uint)` cast on the light count. No behavior change.

## Files
| File | Change |
|---|---|
| `shaders/points/draw/DrawTubes.hlsl` | arc-length UVs, UseWAsU branch, cap disks, warning fixes |
| `shaders/points/draw/DrawTubesArcLength.hlsl` | **new** — arc-length compute pre-pass |
| `Symbols/point/draw/DrawTubes.cs` | +TextureScale, +UseWAsU (−TextureMode enum) |
| `Symbols/point/draw/DrawTubes.t3` | pre-pass nodes, wiring, defaults |
| `Symbols/point/draw/DrawTubes.t3ui` | new input UIs + tooltips for all inputs |
| `.help/docs/.../DrawTubes.md` | full input documentation |


pixtur:
Thanks for this contribution

PR #1165, opened 2026-09-09
feat: new operators RepeatShapePoints and ClosedLoopMesh shape repeater operators

rofoto:
Adds two new operators for After Effects style shape repetition:

- **[RepeatShapePoints]**: repeats a closed shape point loop Count times. Each copy is transformed by its index (OffsetPerCopy / RotatePerCopy / ScalePerCopy, rotation around the shape centroid). All vertices of a copy share Point.F1 = copy index so downstream ops can address individual repeated shapes. Outputs a Point buffer laid out for [DrawClosedLines] plus a ShapeSize output for PointsPerShape wiring.

It works with open shapes too but open shapes become a single connected shape with the repetitions in between the repeater start and end 

- **[ClosedLoopMesh]**: turns closed point loops into a flat, filled triangle mesh via per-plane ear clipping, with planar UVs, per-face normals, ColorRgb and Selection (= source Point.F1). Consumes the same Point buffer / PointsPerShape as [DrawClosedLines].
 

https://github.com/user-attachments/assets/4fd01836-f820-4492-b8fb-79632d223c07

### DIFF
--- a/Operators/Lib/Symbols/mesh/generate/ClosedLoopMesh.cs
+++ b/Operators/Lib/Symbols/mesh/generate/ClosedLoopMesh.cs
@@ -110,15 +110,16 @@ internal sealed class ClosedLoopMesh : Instance<ClosedLoopMesh>
                     Vector3 Pc = positions[tri.Z];
 
                     var faceNormal = Vector3.Normalize(Vector3.Cross(Pb - Pa, Pc - Pa));
-                    if (Vector3.Dot(faceNormal, loopNormal) < 0)
+                    if (faceNormal.Z < 0)
                         faceNormal = -faceNormal;
 
                     AddPlanarVertex(vertices, Pa, loop[tri.X], faceNormal, uAxis, vAxis, plane[tri.X], minU, rangeU, minV, rangeV);
                     AddPlanarVertex(vertices, Pb, loop[tri.Y], faceNormal, uAxis, vAxis, plane[tri.Y], minU, rangeU, minV, rangeV);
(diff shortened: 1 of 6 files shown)


newemka:
[RepeatAtPoints] can already do the same as [RepeatShapePoints].
 

https://github.com/user-attachments/assets/65849ca1-62c6-41e6-b422-a75963c5f31b

[ClosedLoopMesh] is really neat! ♥

image

Please flip the normals in [ClosedLoopMesh] thus it'll be visible by the default camera position.

rofoto:
Normals flipped in [ClosedLoopMesh] 

The flat fill now faces the default camera `(0, 0, +DefaultCameraDistance)`. The orientation is now derived from the camera-facing direction rather than the input loop's winding, so it works consistently whether a shape is authored clockwise or counter-clockwise (verified both against the default camera in the editor, alongside [NGonMesh] for reference).

pixtur:
There is also [RepetitionPoints] that could be combined with [RepeatShapePoints].

I like the idea of the [ClosedLoopMesh] but I have some concerns about the current implementation:
- GPU readback is very(!) tricky when done wrong. It needs proper async buffer reuse to be non-blocking and without excessive memory use. There are few operators doing this (like [PointsToCpu]) and we should avoid replicating the functionality multiple times. It makes this op do something that's not obvious  (GPU->CPU->GPU roundtrip), and will add a frame delay. So unless we find a compute implementation of the triangulation we its side effects are too dangerous. 
- It slightly overlaps with the upcoming geometry ops that support triangulation on CPU side. I have the feeling that 95% of all usecase could be implemented CPU->GPU

PR #1171, opened 2026-09-13
Reroute(Combine) / Cut mouse gestures

H445:
Reroute/Combine: ALT+RMB+Drag
https://github.com/user-attachments/assets/278f8d1e-9570-427f-a399-fd57240386d6

Cut: CTRL+RMB+Drag
https://github.com/user-attachments/assets/06cb7b0f-aab6-4056-81d6-10dad1746a49

Added check to ensure we have no floating single reroute anchors:
https://github.com/user-attachments/assets/f682cd7e-0a48-477b-90fc-010afac6ed95

Drag input/output/shake tests:
https://github.com/user-attachments/assets/1b129dd1-f308-48d4-8701-8240b837bdd0

Multi-type reroute (and undo):
https://github.com/user-attachments/assets/adf70a44-7058-4ebe-a064-8d8d7c9194c9


Editor/Gui/MagGraph/Interaction/ConnectionStroke.cs
```diff
@@ -0,0 +1,316 @@
...
+
+namespace T3.Editor.Gui.MagGraph.Interaction;
+
+/*
+ * Tracks a cut or reroute gesture against the actual tessellated paths drawn by the canvas.
+ * Mouse segments, hit tolerance, and the bounded trail use screen coordinates; stored hits and
+ * anchor positions use canvas coordinates. Occurrences are collected once and committed on release.
+ * _compositionId and _version invalidate the gesture if its graph changes, while reusable buffers
+ * avoid rebuilding placement data on every frame when no additional wire has been crossed.
+ */
```
pixtur, line 15: Please always use an XML comment with a summary block.

Editor/Gui/MagGraph/Interaction/ConnectionStroke.cs
```diff
@@ -0,0 +1,316 @@
...
+internal sealed class ConnectionStroke
+{
+    internal ConnectionStroke()
+    {
+        ObservePath = TestDrawnPath;
+    }
+
+    // Invoke after SetConnection and before PathStroke clears the draw list's path; the delegate is reused.
```
pixtur, line 23: Internal and private fields may use the  /** */ notation as a shortcut for <summary/>

Editor/Gui/MagGraph/Interaction/MagItemMovement.cs
```diff
@@ -57,24 +57,144 @@ internal void PrepareDragInteraction()
         UpdateSnappedConnectionsToDraggedItems();
     }
 
-    internal void CompleteDragOperation(GraphUiContext context)
+    /*
```
pixtur, line 60: So above.  To asterics would be preferable here.

/** 
*
*/

Editor/Gui/MagGraph/Interaction/RerouteOperations.cs
```diff
@@ -0,0 +1,987 @@
...
+        return true;
+    }
+
+    /*
+     * merging requires value-only wiring: a scalar reroute cannot preserve a composition
+     * multi-input bundle or output metadata. Cutting may remove either. plannedChildren supplies
+     * definitions for anchors that the command will create or restore but that are not live yet.
+     */
+    private static bool TryGetConnectionType(Symbol symbol, ConnectionOccurrence occurrence, bool merging, out Type? type, out string error,
+                                             IReadOnlyDictionary<Guid, Symbol>? plannedChildren = null)
```
pixtur, line 538: Would be great to have a nullable annotation here. I.e. NotNullWhen(false)

Editor/Gui/MagGraph/States/GraphUiContext.cs
```diff
@@ -170,31 +172,59 @@ internal bool TryGetActiveInputLine(out MagGraphItem.InputLine inputLine)
...
     {
-        MacroCommand ??= new MacroCommand(title);
-        return MacroCommand;
+        return MacroCommand ?? StartMacroCommand(title);
     }
     
     internal void CompleteMacroCommand()
     {
         Debug.Assert(MacroCommand != null);
+        if (_rerouteCleanupCommand != null)
```
pixtur, line 198: I have the inkling that this could be a method in RerouteOperations

Editor/Gui/MagGraph/States/GraphUiContext.cs
```diff
@@ -254,4 +284,7 @@ public ChangeSymbol.SymbolModificationResults DrawDialogs(ProjectView projectVie
 
         return results;
     }
-}
\ No newline at end of file
+
+    private RemoveDisconnectedReroutesCommand? _rerouteCleanupCommand;
```
pixtur, line 288: Do we really need that? Why couldn't we use the generic MarcroCommand for that?
I thought that was its purpose.

Editor/Gui/MagGraph/Ui/MagGraphCanvas.DrawConnection.cs
```diff
@@ -18,9 +18,20 @@ private void DrawConnection(MagGraphConnection connection, ImDrawListPtr drawLis
         if (connection.Style == MagGraphConnection.ConnectionStyles.Unknown)
             return;
 
+        // Preview geometry converges at the surviving dot; wires between the merging anchors disappear.
+        var sourceInMerge = context.ItemMovement.TryGetRerouteMergePreview(connection.SourceItem, out var mergeSource, out var sourceRadius);
+        var targetInMerge = context.ItemMovement.TryGetRerouteMergePreview(connection.TargetItem, out var mergeTarget, out var targetRadius);
+        if (sourceInMerge && targetInMerge)
+            return;
+
         if (connection.SourceItem.IsCollapsedAway && connection.TargetItem.IsCollapsedAway)
```
pixtur, line 27: What? I'm so confused by this code. It points to a fundamentally different implementation that what I would have had in mind:

I assumption is like this:
1. Slicing process does interaction and results in sets of affected connection lines that could be merged. 
2. For each of them we create a new reroute op
3. We replace all affected connections with on connection into and multiple out of the reroute op.

We could basically emulate result with a [Value] op. So nothing changed in the way connections should be drawn or handled. It this would be the case, what does this code do?

pixtur: Considering that this is a "inner loop" called potentially 1000s of times per frame, I hesitant and worried about the amount of complexity this method got.

Is this overhead justified?

pixtur (changes requested):
Wow. That feature turned out to much more complex than I expected.

Most of my comments are simple formatting preferences related to XML comments.

But the major architectural question is: Can we avoid additional branches in the inner draw loops and implement that feature without polluting so many classes. My hunch was, that _"reroutes" are just ops that are shown differently_.

So _only_ during the interaction (the graph state it active) a "drawer method" could draw everything related the slicing: curve, hovered, intersection points. etc. This would be on top of drawing existing the connection lines which would remain unaffected.

`SymbolLibrary` and `SymbolBrowser` should know nothing about rerouting:

Hiding certain symbols from symbol-browser and symbol lib is an orthogonal task. I would have done this by adding some new property like `HiddenAsSymbol`. This could be relevant for other ops too and could be initialized on load time by `SymbolAnalysis` or another of the already existing helper classes.

Same for snapping. We could introduce a new property like "IsSnapable" (worst name ever). Once again this could be inferred from an `Attribute` on the `Symbol` class (e.g. `[IsNotSnapable]` or `[UiAttributes(NotSnapable= true, HiddenAsSymbol=true)]`.

Besides from this I am really worried about the amount of code. I'm the last one that can blame anybody for this: with AI generated code we tend to implement more complex features and fringe cases because "it's easy". But I'm afraid that complexity is going to bite us later.

I have nothing against long refined self contained methods that draw a beautiful UI interface. But I'm very concerned, if we let the AI start to introduce all kinds of special fields and state variables, like

```
    private Guid _previewDraggedId;
    private Guid _previewTargetId;
    private Guid _previewCandidateId;
    private Guid _previewCandidateDraggedId;
    private int _previewStructureCycle;
    private bool _previewCandidateValid;
```

MagItemMovement is by far one of the most complicated and error prone pieces of code in the code base. I could very well live without merging reroutes (even though I suggested the feature) I we keep complexity and down. Most uses would not miss it or just use shake (a feature that just works, because Reroutes are just ops).

Basically:
- more decoupling (isolating features with interfaces, etc).
- less states

It took me an hour to review the code. I don't wanna give up on that.

I really hope that feedback is not too demotivating. I love the feature and would merge (and refine) the branch anyway.

Editor/Gui/MagGraph/Interaction/ConnectionStroke.cs
H445: Full documentation pass is in progress / will convert all to summary blocks

Editor/Gui/MagGraph/Interaction/ConnectionStroke.cs
H445: these will mostly all be converted to summary blocks too. for intellisense/tools

Editor/Gui/MagGraph/Interaction/MagItemMovement.cs
H445: these will mostly all be converted to summary blocks too. for intellisense/tools as well

Editor/Gui/MagGraph/Interaction/RerouteOperations.cs
H445: Added [NotNullWhen(true)] to out Type? type, since the non-null guarantee applies when the method succeeds. This also lets us remove the caller’s !. The parameter contract is now documented as well.

Editor/Gui/MagGraph/States/GraphUiContext.cs
H445: Moved this into RerouteOperations.CompleteCleanup. GraphUiContext now delegates cleanup and refreshes its layout when needed. Cleanup remains last in the undo group so undo restores reroute nodes before their connections.

Editor/Gui/MagGraph/States/GraphUiContext.cs
H445: We do use MacroCommand for grouped undo/redo. This field holds pending cleanup and records which reroutes were connected before the edit, so cleanup only removes nodes that became disconnected during that edit. Cleanup is appended last so undo restores the nodes before their wires. I moved completion into RerouteOperations.CompleteCleanup; the macro still owns undo/redo.

Editor/Gui/MagGraph/Ui/MagGraphCanvas.DrawConnection.cs
H445: - The confusing code previewed dragging two existing reroutes together by hiding one dot and redirecting their wires. I removed that feature and its preview logic.

- Profiling 1,000 reroutes led to two allocation fixes: removing per-wire closures and reusing layout sorting storage. Wire drawing now measures zero managed allocations; graph-pass allocations fell by 83%.
- I was able to bring down this overhead quite a bit. 
image

H445:
Yet another documentation pass.

Removed redundant private/internal summaries; retained concise comments for non-obvious behavior.
Removed 522 unnecessary parameter documentation entries.
Removed all comments from all 78 routing operators.

H445:
The last doc pass was way too aggressive. I've reverted all previous contributor comments (git blame etc) and only modified where necessary.

